PR #16, opened 2015-09-03
Removed redhat_release role & its references

afomm:
- To avoid going through the redhat_release role many times

Signed-off-by: Afom Michael <tmichael@redhat.com>


cevich:
Okay! Looks good to me.

PR #29, opened 2016-03-23
Migrate most of CI-level operations here

cevich:
Almost working


site.yml
```diff
@@ -1,4 +1,48 @@
...
-- include: redhat_release.yml
-- include: autotested.yml
-- include: cleanup.yml
+
+# Setup/override global variables
+- hosts: all
+  tasks:
+    # Defines adept_job_no_group and rt_adept_job_no
+    - include: roles/common/tasks/adept_job_no.yml
+
```
edsantiago: What would you think of:

    vars:
      some_descriptive_variable_name: "{{ groups[adept_job_no_group] }}"

site.yml
```diff
@@ -1,4 +1,48 @@
...
+  tasks:
+    - debug: var=adept_job_no_group
+
+- hosts: peons
+  roles:
+    - role: peon
+      # During cloud-type cleanup, peons will be deleted
+      when: adept_context == "cleanup"
+      # In case other jobs are running concurrently
+      with_items: "{{ groups[adept_job_no_group] }}"
```
edsantiago: ...instead of just "{{ item }}" ?

site.yml
cevich: I considered that, however it has a higher precedence (http://docs.ansible.com/ansible/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable) and I need to be able to override the place-holder value (by a ``set_facts``).  It's kind of a subtle "gotcha" that's bitten me before, so I prefer to always go for lowest-precedence possible.  In fact, there are some items in ``files/variables.yml`` that I should move to ``group_vars/all.yml``.

Also, ``site.yml`` is kind of just an "example", I'd like to keep most things modular so they can be plugged in different ways in alternate play-books depending on the situation/environment.

site.yml
cevich: ohh, that reminds me, I should probably not use ``item`` as the loop variable (it can clash with deeper/nested loops)...

site.yml
cevich: Argh, actually might need to do something like this.  I just predicted and proved there are problems at the top-level if the variable isn't available at the playbook scope.  Uggg.   Still, it's two-steps forward, and only one back :smile:   I'll figure something out.

cevich:
@edsantiago This is "working" now (at a basic level) w/ openstack dynamic inventory and there shouldn't be any management-clashes over concurrent runs.  What I'm doing to make testing easier is, run adept against this yaml instead of ``files/slave_container.yml``:

```diff
--- files/slave_container.yml	2016-05-12 16:43:42.583109799 -0400
+++ files/slave_private.yml	2016-05-12 16:46:52.227711039 -0400
@@ -4,7 +4,7 @@
 - command:
     filepath: /usr/bin/cp
     arguments: |
-        $HOME/.ssh/id_rsa $WORKSPACE/ssh_private_key
+        $HOME/.ssh/__INSERT_RSA_KEY_UPLOADED_TO_OPENSTACK_
 
 # Runs anywhere w/ only python interpreter, docker, and this repo
 - command:
@@ -39,6 +39,10 @@
             # TODO: Read-only when --tmpfs available in 1.10+
             # --read-only
             # Image built above
+            -e OS_AUTH_URL=__INSERT_QUOTED_URL__
+            -e OS_USERNAME=__INSERT_QUOTED_USERNAME__
+            -e OS_PASSWORD=__INSERT_QUOTED_PASSWORD__
+            -e OS_TENANT_NAME=__INSERT_QUOTED_USERNAME__
             adept_slave:latest
             $ADEPT_CONTEXT
             /var/lib/workspace
@@ -47,28 +51,6 @@
             # see last item (below)
             $ADEPT_OPTIONAL
 
-- command:
-    contexts:
-        - cleanup
-    pass_context: False
-    filepath: /usr/bin/bash
-    # N/B $ADEPT_OPTIONAL used to pass non-default image max lifetime
-    arguments: |
-        -c 'set -e;
-
-            # Three-days by default
-            MAX=259200;
-
-            CREATED=$(docker inspect --type=image \
-                      --format="{{.Created}}" adept_slave:latest);
-            AGE=$[$(date +%s) - $(date -d $CREATED +%s)];
-
-            [ $AGE -lt $MAX ] || \
-            # Do not prune potentially still viable build-cache
-            /usr/bin/docker rmi --force \
-                                --no-prune \
-                                adept_slave:latest;'
-
 # Don't leave private keys hanging around
 - command:
     filepath: /usr/bin/rm
```

Oh, also don't forget to change ``files/variables.yml`` ``openstack.key_name`` to the name you gave when you uploaded your key :grinning:

edsantiago:
Sob. I keep understanding less and less. So on one side this eliminates the need to hand-edit `files/variables.yml` because the env lookups should now work, but I still need to edit the file for key_name, and now have to hand-edit slave_private.yml? I keep thinking there has to be a simpler way...

cevich:
> hand-edit files/variables.yml

This is the current dumping ground for all/most important variables,  having to edit/change this is going to be quite common.

> and-edit slave_private.yml

OIC, this is wonky because I'm lazy and like to use the container to access the openstack command-line clients on occasion.  I probably should just have a separate container for that :smile:

Jenkins is a bit easier to setup to bring in creds. by env-vars, that's why those are there.  
But for manual runs, it's fine to stick them right into ``files/variables.yml`` if that's easier for you.

Another option is ``files/openstack.yml`` or indirection through ansible-vault.  In either case, I just wanted to avoid accidentally checking them into the repo. :smile:

cevich:
Oh, you also need to rebase from ``upstream/master`` as there are changes.

cevich:
Oh right, the other annoyance is, the silly ``openstack.py`` dynamic inventory doesn't take it's creds from ansible, it uses either ``the files/openstack.yml`` or env. vars., which need to be set from the parent process.

One idea is I could add a call to ansible from ``.bash_profile`` inside the container, and then export from there.  That way they can be centrally located in ``files/variables.yml`` and allow our hair to grow a bit longer w/o so much yanking :smile:

cevich:
Last push squashes a bunch of commits and consolidated all configuration input into ``files/variables.yml``.  Successfully provisions example centos peons, installs packages and updates.  Cleanup also properly removes all created systems.

cevich:
jenkins, test this please.

cevich:
jenkins, test this please.

cevich:
jenkins, test this please.

PR #38, opened 2016-08-17
Workarounds, updates, and features.

cevich:


roles/autotested/templates/subtests.ini.j2
```diff
@@ -15,7 +15,7 @@ subsubtests = good,wrong_tag,wrong_registry_addr
 config_version = 0.8.5
 generate_generic = yes
 ping_url = "www.redhat.com"
-cmd = 'ping -c 3 -q %(ping_url)s'
+cmd = 'yum install -y iputils && ping -c 3 -q %(ping_url)s'
```
edsantiago: This never works for me: my docker containers can't run yum. Does it work for you?

cevich: Ya, if the host is registered it should work, mostly, well, depends on what it's registered too :grinning:

cevich: or the host is centos atomic.

PR #49, opened 2016-08-30
Ignore Autotest execution errors

cevich:
Otherwise results (good/bad) will not be collected for debugging.


cevich:
Yay! Only //TWO// test failures now.  Amazing what timeout fixing can do.

    * autotest.docker/subtests/docker_cli/commit.1
    * autotest.docker/subtests/docker_cli/cp.2

Calling that good 'nuf.

PR #51, opened 2016-09-06
Fix Autotest/Docker Autotest cache clash

cevich:
The autotested role first transfers a copy of the autotest repository to
the test VM.  If recursive submodules were used to clone that repo,
it will include a (likely very) old version of docker autotest.  Later,
when the cached (modern) docker-autotest is also transfered, it will end
up with bits that have long since been removed.

Fix this by first removing the docker autotest path before doing the
cache copy.

### DIFF
--- a/roles/autotested/tasks/main.yml
+++ b/roles/autotested/tasks/main.yml
@@ -32,6 +32,11 @@
     path: "{{ autotest_dest_dir }}/RELEASE-VERSION"
     state: absent
 
+- name: Bundled Docker Autotest directory is removed
+  file:
+    path: "{{ autotest_docker_path }}"
+    state: absent
+
 - name: Cached Docker Autotest is copied
   synchronize:
     dest: "{{ autotest_docker_path }}"


roles/autotested/tasks/main.yml
```diff
@@ -32,6 +32,11 @@
     path: "{{ autotest_dest_dir }}/RELEASE-VERSION"
     state: absent
 
+- name: Bundled Docker Autotest directory is removed
+  file:
+    path: "{{ autotest_docker_path }}"
+    state: absent
+
```
edsantiago: FWIW here's my playbook

     # git repos: autotest and autotest-docker. update=no means "chill": it
     # lets us modify files in the repos without Ansible barfing on future
     # runs. If you ever want to update, just rm -rf the checkout manually.
     - name: install autotest
        git: repo=https://github.com/autotest/autotest.git
            dest={{ destdir }} update=no recursive=no

     - name: install autotest-docker
       git: repo=https://github.com/autotest/autotest-docker.git
            dest="{{ destdir }}/client/tests/docker" update=no

cevich: Unf. the clone happens in Jenkins before ADEPT is even called.  That's needed so that the pull-request triggering and change-tracking can work.

However, what you have above would be a great fall-back to add in, e.g. when cache (jenkins repo clone) isn't available.  That would let ADEPT run more stand-alone, not requiring pre-cloned autotest/docker autotest repos.

cevich:
@edsantiago okay, opened issue #52 so I don't forget your idea.

PR #58, opened 2016-10-03
Fix build/https_file on RHEL* VMs

cevich:
This test relies on building from a remote, https based docker file.
The test file uses the CentOS base image, so rather than use a different
url, just patch up the configuration for the extra build pull.


cevich:
Uggg, big giant python traceback from travis, something about more than 100 named groups.  WTF?

cevich:
Red Hat Internal CI bustid ATM.  Got to fix that first before this goes in.

cevich:
This is not necessary anymore.

PR #63, opened 2016-11-07
Fix build/git_path, Role deps, Linear Strategy, Atomic Storage Reset

cevich:
Mostly all niggly-details.  The ``build/git_path`` one is most important.

### DIFF
--- a/group_vars/openstack
+++ b/group_vars/openstack
@@ -28,20 +28,20 @@ adept_openstack:
     # Mapping of image name + version to openstack image uuids
     image_map:
         centos:
-                "7.2": "3282ac5a-8b7e-4dc6-a36d-b13a276638f0"
+                "7.2": "CentOS-7-x86_64-GenericCloud"
         centosatomic:
-                "7.2": "dbd01b32-3a5c-4bcb-9c4b-2857d17b4567"
+                "7.2": "CentOS-Atomic-Host-7-GenericCloud"
         rhel:
-                "7.2": "e14d0aa8-17fb-45c3-b842-056d987acec9"
+                "7.2": "rhel-7.2-server-x86_64-released"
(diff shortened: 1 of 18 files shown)


roles/autotested/templates/control.ini.j2
```diff
@@ -17,13 +17,17 @@ include =
 exclude = example,subexample,pretest_example,intratest_example,
           posttest_example,
           docker_cli/run_volumes,
-{% if not is_atomic %}
+{% if is_atomic %}
+          # Atomic host does not have git
           docker_cli/build/git_path,
-    {%- if is_enterprise %}
+
+    {%- if not is_enterprise %}
```
edsantiago: I don't get this. Isn't debuginfo applicable on both RHEL and Atomic?

cevich: Struggling, but IIRC, there was something about the test that doesn't work on Atomic.  Nope, it's gone, don't remember.  I can try and enable it and we can see what happens...

roles/autotested/templates/control.ini.j2
```diff
@@ -20,11 +20,6 @@ exclude = example,subexample,pretest_example,intratest_example,
 {% if is_atomic %}
           # Atomic host does not have git
           docker_cli/build/git_path,
-
-    {%- if not is_enterprise %}
-          # Only applicable to RHEL Atomic
-          redhat/packaging,
-    {%- endif %}
```
edsantiago: 584cb45 perhaps? Test was failing on Atomic because of missing `magic.py`, this commit fixed it.

group_vars/openstack
```diff
@@ -28,20 +28,20 @@ adept_openstack:
     # Mapping of image name + version to openstack image uuids
```
edsantiago: comment needs fixing too

PR #67, opened 2016-11-10
Barrel-o-fixes

cevich:
Mostly Fedora-support related + one special request

### DIFF
--- a/group_vars/fedora/installed_rpms.yml
+++ b/group_vars/fedora/installed_rpms.yml
@@ -4,7 +4,6 @@ use_dnf: True
 
 installed_rpms:
     - docker
-    - docker-latest
     - atomic
     - docker-selinux
     - procps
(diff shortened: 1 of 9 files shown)


cevich:
Note: There's a big SNAFU I'm attempting to fix with 'Fix action/state change on non-extent services'.  For some reason ansible's service module pukes if you tell it to turn off/disable a service that doesn't exist.  I've got it correctly filtered, but something is blocking the task from running when it should be...still debugging.

cevich:
Gaarrr, still broken.  Setup job should have restarted docker service on centos VM.

cevich:
There we go, got it now. One more run to confirm then I'll merge.

cevich:
...patiently waiting...

cevich:
last try :confounded: 

(updated / installed role not setting ``needs_reboot`` for some reason, try to fix that and reload systemd - docker service isn't listed)

PR #68, opened 2016-11-14
Missing input dir arg in results2junit call

edsantiago:


cevich:
LGTM.  I've got it pulled into the devel branch so I can do a test-run on it and another fix for the services/rebooting problem.

PR #69, opened 2016-11-14
Devel

cevich:


cevich:
OOOooooohhhhh!    I got it!  have to use ``systemctl ... list-unit-files ...`` to see all installed units, not just the enabled ones.  THAT's the problem all along.  Fixing...

PR #70, opened 2016-11-15
Pass peon metadata name into results2junit

cevich:
Test VMs (peons) are all assigned a globally unique name to prevent
clashes with concurrently running jobs.  This is not a good name to use
for tagging test results because they will never align from run-to-run.

This change makes the peon and cloud metadata available to them via
host_vars.  Later, when the autotest role is applied, it's possible to
extract the peon's uncompounded (basic definition) name.  This is
guaranteed unique within a run, but will be (mostly) shared across
them.

### DIFF
--- a/roles/autotested/tasks/produce_results.yml
+++ b/roles/autotested/tasks/produce_results.yml
@@ -10,13 +10,22 @@
         autotest_execution is defined and
         autotest_execution|success
 
+- debug:
+    msg: "I am {{ peon_metadata.name }}"
+  when: peon_metadata is defined
+
+# Extract a name unique to this run, but common across runs of same job
 - name: Attempt to run alternate results2junit
-  shell: "{{ autotest_docker_path }}/results2junit --name {{ inventory_hostname }} results/default > {{ autotest_dest_dir }}/client/results/default/results.junit"
+  shell: >
+    {{ autotest_docker_path }}/results2junit
+        --name {{ peon_metadata.name }}
+        results/default
   args:
     chdir: "{{ autotest_dest_dir }}/client"
   register: results2junit
   ignore_errors: true
-  when: docker_autotest.execute and
+  when: peon_metadata is defined and
+        docker_autotest.execute and
         autotest_execution is defined and
         autotest_execution | success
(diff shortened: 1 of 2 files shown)


cevich:
[(?) [CI/CN] SPIKE: HOWTO: De-fubar jUnit results](https://trello.com/c/c9wTyspx/323-ci-cn-spike-howto-de-fubar-junit-results)

roles/autotested/tasks/produce_results.yml
```diff
@@ -10,13 +10,23 @@
...
+  when: peon_metadata is defined
+
+# Extract a name unique to this run, but common across runs of same job
 - name: Attempt to run alternate results2junit
-  shell: "{{ autotest_docker_path }}/results2junit --name {{ inventory_hostname }} results/default > {{ autotest_dest_dir }}/client/results/default/results.junit"
+  shell: >
+    {{ autotest_docker_path }}/results2junit
+        --name {{ peon_metadata.name }}
+        results/default
+        > {{ autotest_dest_dir }}/client/results/default/results.junit
```
edsantiago: I've fixed my script so it writes directly to results.junit in the same directory that's given as input arg. I will do the same in the python rewrite. So you can probably remove line 23 entirely.

roles/openstack/tasks/setup.yml
```diff
@@ -18,25 +18,35 @@
...
+
   rescue:
     - include: roles/openstack/tasks/cleanup.yml
       vars:
         # Help out with debugging
         cleanup_peon_facts: False
         cleanup_clouds_yaml: False
   when: peon_facts is undefined
+
+- name: peon's metadata written into it's host_vars
```
edsantiago: No apostrophe in 'its' :)

roles/autotested/tasks/produce_results.yml
cevich: excellent

roles/openstack/tasks/setup.yml
cevich: doh!

PR #72, opened 2016-11-17
Fix ignored-failure on atomic upgrade

cevich:
Also, give a known fixed size for root storage on atomic hosts and add beginnings of feature to support detection of partially-setup hosts.


cevich:
[(1) [CN] Fix ADEPT setup continuing despite failure](https://trello.com/c/8H2frKKH/333-1-cn-fix-adept-setup-continuing-despite-failure)

PR #76, opened 2016-11-23
Allow autotest client to fail

cevich:

### DIFF
--- a/roles/autotested/tasks/main.yml
+++ b/roles/autotested/tasks/main.yml
@@ -114,7 +114,7 @@
       args:
         chdir: "{{ autotest_docker_path }}"
       register: autotest_execution
-      async: 14401
+      async: "{{ (docker_autotest_timeout | default(240)) * 60 }}"
       poll: 60
       # Any Autotest failures will eventually be reflected as missing results
       ignore_errors: True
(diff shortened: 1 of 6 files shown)


edsantiago:
Just a question on the commit message: could you expand on why? as in, what problem you see with d-a-t failing, and how this change addresses it, and how failure will be handled?

cevich:
> Just a question on the commit message: could you expand on why? as in, what problem you see with d-a-t failing, and how this change addresses it, and how failure will be handled?

I wish I remembered why I did this originally.  There must have been a case where a non-zero exit caused some "worse thing" to happen.  I can't imagine d-a-t exiting non-zero on test failure, but that may be the reason.  When CI finishes, we'll know for sure then I can update the commit message :grinning:

cevich:
OIC.  Commit 191fa7d5 says why.  Yeah, there's a better way to do this...

roles/autotested/tasks/main.yml
```diff
@@ -107,19 +107,24 @@
...
+        chdir: "{{ autotest_docker_path }}"
+      register: autotest_execution
+      async: 14401
+      poll: 60
+      # Any Autotest failures will eventually be reflected as missing results
+      ignore_errors: True
 
-- include: produce_results.yml
-  when: autotest_execution is defined
+  always:
```
cevich: works just like an ``except...finally`` in python, except it doesn't re-raise the error (in ansible)

PR #78, opened 2016-11-29
New adept.py features

cevich:
* Rename adept ``yaml-file`` -> transition file, ``.xn`` extension.
This provides a more clear distinction from Ansible playbooks,
and more accurately describes their purpose: Orchestration
steps necessary to transition from one system-context to another
(i.e. setup -> run -> cleanup).

* Beefed up "environment protection" with a list of pre-approved
environment variable names.  In particular, this prevents leaking of
credentials or other sensitive information into Ansible.

* Removed hard-coding of ssh private key name/location.  This is safer
to pass in via env. vars, and/or ``ansible.cfg``.  Allows flexibility,
and provides more clarity in purpose when multiple keys are used/needed.

* Added new ``variable`` action-item, which allows setting of global,
transition-scoped keys and values.  Their values can be set directly, or
by indirection from another named env. var, or by reading from a named
file.

* Added just-in-time resolution of global variables, into action-item
specific references.  The variables are handled separately by the
particular action being taken.  For example, the ``playbook`` action
will pass global variables in via the ``--extra-args`` command line
parameter.  With the ``command`` they are sanitized and passed in as
env. vars.  All variable values are resolved (looked up or set) just
prior to their use.  This enables, e.g., playbooks or commands
themselves to change or set values dynamically.

* Updated and added new unittests and comments for all changes and
features listed above.


cevich:
[(1) (CN) adept.py variable translation](https://trello.com/c/g81uGycX/275-1-cn-adept-py-variable-translation)

PR #79, opened 2016-11-29
Fix atomic_upgrade and build/git_path template comments

cevich:
Simplify failed/changed conditions, guarantee the _needs_upgrade is set
in local scope, and don't check atomic_host_upgrade if _needs_upgrade is
false. This fixes a problem where there is no upgrade available, causing the
upgrade block to be bypassed. This means the atomic_host_upgrade is
undefined, though referenced after the block.


cevich:
Moved just these changes from #78 which is all new-features

PR #92, opened 2017-01-18
Quick-docs

cevich:
Something is better than nothing.


cevich:
moved to https://github.com/cevich/autotest-docker-enabled-product-testing/pull/101

PR #95, opened 2017-01-23
Restore copy-files feature of autotested role

cevich:
This is needed where files/directories need to be transfered to peons
after Autotest/Docker Autotest is transfered, but before any automatic
configuration is generated.  This provides for an oportunity to override
the automatic/default configurations and/or provide additional
non-standard content (such as private tests).

Resolved the long-standing TODO of making the autotested role
context-aware.  In the grand-scheme of ADEPT, this allows preservation
of peon setup/configuration state prior to actually running any tests.
It also makes it easier to perform extra steps outside of ADEPT in between
'setup' and the beginning of the 'run' transition.

Finally, move the template list outside of the docker_autotest dict
into it's own top-level variable.  This makes it much easier to override
from higher scoped group or host_vars.  i.e. otherwise, those scopes
would have to redefine the entire docker_autotest dict.


edsantiago:
Oh, wow, this is enormous! I put the new file into dockertest because I remember you saying it would be trivial. Before we go too far down this path, is there a way to get `known_failures.csv` into the ADEPT runs that really is trivial?

cevich:
> Oh, wow, this is enormous! 
> ...
>I remember you saying it would be trivial.

Trivial for dockertest (RH internal repo) yes (and I just pushed that up).  Here, I totally forgot that I had yanked out the copy feature long ago.    Then when I pulled it back in, I remembered why I pulled it: It's a PITA to use (requires overwriting the entire ``docker_autotest`` variable.

So figured I might as well fix that TODO anyway...and...and...eek rolling snowballs!

Okay, lemmie see about breaking this down into more distinct commits, that should help a bit...

cevich:
@edsantiago okay there, that's much cleaner and I also fixed two bugs.  Waiting for CI results to see if there are any more gremlins.

group_vars/fedoratomic/autotested.yml
```diff
@@ -11,11 +17,6 @@ docker_autotest:
         name:
         tag:
         build_url:
-    templates:
-        - src: "{{ custom_subtests_ini_j2 | default('subtests.ini.j2') }}"
-          dest: "{{ autotest_docker_path }}/config_custom/subtests.ini"
-        - src: "{{ custom_control_ini_j2 | default('control.ini.j2') }}"
-          dest: "{{ autotest_docker_path }}/config_custom/control.ini"
     preserve_fqins:
         - "docer.io/fedora:latest"
```
edsantiago: Completely unrelated to your change, but shouldn't this be docker?

edsantiago:
I can't review this. Sorry, it's much more complicated than my understanding of ADEPT. I'll trust your judgment on whether or not to check it in or find a simpler way. All we're trying to do is take a file from one place and put it in the working directory. It's a little alarming that it takes so much code and rearranging and duplication to do that.

group_vars/fedoratomic/autotested.yml
cevich: Ahh, woops, missed that one.  Should be an empty-list.  Okay, fixed.

cevich:
> I can't review this. Sorry, 

No worries.  I'll give it a few more goes, and if I can't hack it...I'll do what  my little bobble-head named "Ed" keeps reminding me: "do it simpler...make it easy...pretty code is for wimps..." 😀

cevich:
```
15:50:20 TASK [openstack : service_catalog is stored locally] ***************************

15:50:20 An exception occurred during task execution. To see the full traceback, use -vvv. The error was: RepresenterError: cannot represent an object: endpoints

15:50:20 fatal: [localhost]: FAILED! => {"failed": true, "msg": "Unexpected failure during module execution.", "stdout": ""}
```

Hrrmmm, same error as 79b4999 was suppose to "fix" (in a different place).  So it's official, Ansible is somehow completely loosing it's brains.  I've never seen this error before and it's not coming from any code touched by any commit here....hmmmm....a mystery!

cevich:
Okay, so #96 proves this Ansible error has nothing to do with the changes in this PR.  Problem must already be on master or something has completely broken with the current Kommandir VM and/or container image.

edsantiago:
Could that be why last night's full run failed to run on any RHEL?

cevich:
> Could that be why last night's full run failed to run on any RHEL?

Naa, sent ya a mail about that.

cevich:
Next problem:

```
13:36:56 TASK [autotested : fail] *******************************************************

13:36:57 fatal: [centosatomic-latest-7-2-ADEPT-ci-147-1d93528c275b]: FAILED! => {"changed": false, "failed": true, "msg": "Required file /var/lib/workspace/variables.yml does not exist"}

13:36:57 fatal: [centosatomic-7-2-ADEPT-ci-147-6b92111fa448]: FAILED! => {"changed": false, "failed": true, "msg": "Required file /var/lib/workspace/variables.yml does not exist"}
```

That's probably a goof I've made here, investigating...

cevich:
Okay, that one's fixed, now:

```
"The conditional check 'not is_atomic and (lvg_ops != [] or lvol_ops != [])' failed. The error was: error while evaluating conditional (not is_atomic and (lvg_ops != [] or lvol_ops != [])): 'is_atomic' is undefined\n\nThe error appears to have been in '/var/lib/adept/roles/lvm/tasks/main.yml': line 9, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n# Must happen here, role applied before os-type roles\n- include: roles/installed/tasks/main.yml\n  ^ here\n"
```

Hmmmm, ``is_atomic`` is defined by common role, so now I'm missing a dependency somewhere...

cevich:
Bahhhh, there it is....simple typo was the problem all along 😖

cevich:
(N/B: I'm aware this PR is looooooooooooooong needing to be split between features and fixes, just wanting to get it all working normally again before tackling that)

edsantiago:
That doesn't strike me as a typo; it seems like a point of confusion, and a very easy one to make.

cevich:
> and a very easy one to make.

No kidding!  Weeee hates itsssssss burnssss us itsss burnsss sus! -gollum-gollum-

Okay, all the fixes moved over to: #99   Stripping them out from here...

cevich:
Okay, all the features moved to #100 closing this.

PR #98, opened 2017-01-24
Workaround: Ansible issue 15915

cevich:
Link: https://github.com/ansible/ansible/issues/15915

Error: RepresenterError: cannot represent an object: <foo>

Add comment about forced use of 2.1.0 until 2.1.2+ is available.

Also remove a bunch of python3-foo packages that never existed
(copy-paste error) and rename python-jsonpatch -> python2-jsonpatch (the
former has been deprecated).


cevich:
(already tested in #95)

PR #99, opened 2017-01-25
Barrel-o-fixes

cevich:


files/slave.dockerfile
```diff
@@ -2,21 +2,20 @@ FROM docker.io/stackbrew/centos:latest
 MAINTAINER Chris Evich <cevich@redhat.com>
 
 RUN yum update -y && \
-    yum install -y findutils epel-release https://rdoproject.org/repos/rdo-release.rpm && \
+    yum install -y epel-release https://rdoproject.org/repos/rdo-release.rpm && \
```
edsantiago: Why are you removing findutils? I wonder if this is what's making `run_exec/exec_pid_count` fail on fedora?

cevich: Nope, I just moved it into ``files/slave.rpms``.  This won't affect the test VMs, remember slave == kommandir-container (and needs to be renamed).  i.e. when everything is up and going, the topology looks like this:

```
executing node (e.g. Jenkins) -> kommandir vm -> kommandir container -> peons (test VMs)
```

cevich:
@edsantiago here...maybe I better just annotate the diffs for ya...

files/slave.dockerfile
cevich: Moved ``findutils`` into ``slave.rpms`` because that's where it should be.

files/slave.dockerfile
```diff
@@ -2,21 +2,20 @@ FROM docker.io/stackbrew/centos:latest
...
     # Keep every layer as clean (i.e. small) as possible
     yum clean all
 
-# N/B: Use Ansible 2.1.0 until 2.1.2 is available b/c of:
-#      https://github.com/ansible/ansible/issues/15915
-
```
cevich: Reverted, we're going the other way now - rpm version of 2.2.1 to source version of 2.2.1 (which is newer)

files/slave.dockerfile
```diff
@@ -2,21 +2,20 @@ FROM docker.io/stackbrew/centos:latest
...
 
 RUN xargs -a /root/slave.rpms yum install -y && \
+    # Keep every layer as clean (i.e. small) as possible
     rm -f /root/slave.rpms && \
     rm -rf /usr/src /usr/share/doc && \
     yum clean all
 
-RUN yum install -y python-pip python-devel && \
+# Development jibblets not needed at runtime, allow later removal
+RUN yum install -y python-pip python-devel openssl openssl-devel libffi libffi-devel; \
```
cevich: openssl and libffi are needed to build/install ansible from source

files/slave.dockerfile
```diff
@@ -25,13 +24,24 @@ RUN yum install -y python-pip python-devel && \
     yum groupinstall -y 'Development Tools' && \
     # Very specific version is required for older Openstack Env.
     pip install --no-cache-dir --disable-pip-version-check shade==1.12.1 && \
+
+    # Use Ansible from source until rpm available w/ fix for:
+    # https://github.com/ansible/ansible/issues/15915
+    pip install --upgrade setuptools && \
```
cevich: The ``python-setuptools`` installed from  ``slave.rpms`` is too old, temporarily bump it up using pip (eek!). Effectively, the install in ``slave.rpms`` is just to make sure as many dependencies come in rpm-format as possible (so pip ignores them).

files/slave.dockerfile
```diff
@@ -25,13 +24,24 @@ RUN yum install -y python-pip python-devel && \
...
+    git clone https://github.com/ansible/ansible.git \
+        --branch=stable-2.2 --recursive --depth=0 --single-branch \
+        /usr/src/ansible && \
+    cd /usr/src/ansible && \
+    make install && \
```
cevich: This actually ends up calling pip to do more dependency fullfilment, but hopefully we satisfied most of them via rpms (above).

files/slave.dockerfile
```diff
@@ -25,13 +24,24 @@ RUN yum install -y python-pip python-devel && \
...
+    
     # Keep every layer as clean (i.e. small) as possible
-    yum history -y undo $(cat /root/trans_num) && \
+    yum history -y undo $(cat /root/trans_num) ; \
```
cevich: This keeps the kommandir-container (aka slave) image small.  There's no need to hang onto all the development bits needed to build ``andible`` and ``shade``.  Containers based on this image are spun up and removed frequently, so keeping the image small improves performance substantially.

files/slave.dockerfile
```diff
@@ -25,13 +24,24 @@ RUN yum install -y python-pip python-devel && \
...
     rm -f /root/trans_num && \
     yum clean all && \
     rm -rf /usr/src /usr/share/doc
 
-RUN echo "" > /etc/ansible/hosts && \
+RUN mkdir -p /etc/ansible && \
+    echo "" > /etc/ansible/hosts && \
```
cevich: The previous rpm-based install of ansible too care of making this directory, now we gotta do it here instead.

group_vars/fedoratomic/autotested.yml
```diff
@@ -16,8 +16,7 @@ docker_autotest:
           dest: "{{ autotest_docker_path }}/config_custom/subtests.ini"
         - src: "{{ custom_control_ini_j2 | default('control.ini.j2') }}"
           dest: "{{ autotest_docker_path }}/config_custom/control.ini"
-    preserve_fqins:
-        - "docer.io/fedora:latest"
+    preserve_fqins: []
```
cevich: woopsie

roles/yumrepos/meta/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+
+dependencies:
+  - role: common
```
cevich: explicit dependencies are always better than implicit 😀

vars/peon_defaults.yml
```diff
@@ -54,23 +54,6 @@ peon_defaults:
...
-    fedoratomic:
-      version: "24"
-      groups:
-        - fedoratomic
-        - atomic_upgrade
-        - autotested
-        - peons
-
-    fedora:
-      volume: "20"
```
cevich: Technically we could do both 24 and 25 - it just requires changing the names (``fedoratomic -> fedoratomic24 and fedoratomic25``).  But why bother, easier all around to only test one Fedora.

cevich:
Yay!  EEEET VERKS! EEEEER VERKS!    (no more ``to_yaml`` or unknown var: ``adept_path``))

cevich:
Ha!  Only one subtest failure to boot!  Okay, merging...

PR #101, opened 2017-02-02
Reimplement ADEPT for simplicity and lower viscosity

cevich:
DONE:

*  Reduce dependencies on both packaging and jenkins
*  Remove need for a kommandir container
*  Reduce size & number of ``adept.py``'s ``xn`` files
*  Re-write host provisioning to simply call a specified script (to be written)
*  Reduce number of preparation steps (before executing ``adept.py``
*  Support moving "job" details out of RH Internal Jenkins files and into this repo.
*  Support concurrent (separate process environment) job execution with proper locking (TBD)
*  Support local (i.e. laptop) execution (assuming provisioning script dependencies are met)
*  Migrate the old peon setup (just sitting in the repo now) to the new ``inventory/host_vars/name.yml`` way.
*  Migrate the internal ``full`` and ``basic`` jenkins-centric crap into this repo (``jobs`` dir).
*  Minor updates/tweaks to the peon setup/install and autotested roles.
*  Remove all the disused junk playbooks.
*  Write a proper ``jobs/travis_ci/job.xn`` and update ``.travis.yml`` so it can actually do a trial-run.

Still TODO:
*  [Documentation](http://autotest-docker-enabled-product-testing.readthedocs.io/en/simplify/README.html)
*  Source shared ``exekutir`` and ``kommandir`` roles under a "common" directory or simply combine all the roles under one subdirectory.
*  Rename top-level exekutir playbooks (The sequence-based names are confusing IMHO).  Better to go with something closer to how they're named under ``kommandir``.


cevich:
[(2) (CN) ADEPT Slave Simplification](https://trello.com/c/CevYDUei/279-2-cn-adept-slave-simplification)

cevich:
[(5) (CN) ADEPT's VM creation rewrite](https://trello.com/c/MMH7mHUR/278-5-cn-adept-s-vm-creation-rewrite)

cevich:
[[CN](3) ADEPT QuickStart Docs](https://trello.com/c/OqOJD3gK/81-cn-3-adept-quickstart-docs)

cevich:
[[epic] [CN] Simplify Dockertest/ADEPT](https://trello.com/c/BYE12CIh/266-epic-cn-simplify-dockertest-adept)

cevich:
@afomm @edsantiago Reached a milestone here, but I'm afraid some of this may require more advanced ansible-grok.  However, I'd like to help with that, and I'm also interested in your feedback and contributions.  

What's the best way forward?

Should we have some conference/demo time, would you prefer to look it over on your own first, or something else?

afomm:
@cevich Great! I think a demo/conference would be great. We can also play with it on our own until then.

edsantiago:
I've just skimmed the diffs: it's too much for me, and I think it would be even if I were an ADEPT expert (which of course I'm not). I'm going to skim again looking for just gestalt, but I'll admit that I cannot give this an acceptable review. What I _can_ try is, once it's committed, to start working (again) on understanding the ADEPT flow.

.gitignore
```diff
@@ -2,9 +2,18 @@
...
-hosts
 workspace
-files/*private*
+
+!__exekutir__(/.*)?/inventory/hosts
+
+jobs/.*/.*
+!jobs/.*/.empty
+
+__executir__/host_vars
```
edsantiago: Should this be k (exekutir, not executir)?

I feel distinctly uncomfortable with the odd names like kommandir and exekutir and whatnot. They are too easy to misspell, and much too easy for misspellings to go uncaught: to me and a few other people, misspellings almost-literally jump out of the page. Our eye detects them automatically, effortlessly. This makes it easy to skim code looking for typos. That only works with familiar words.

cevich:
> I've just skimmed the diffs: it's too much for me

@edsantiago don't bother with the diffs, things have moved around so much and large swaths re-written.  Best to just look at the actual code/playbooks themselves.  ATM, it should actually run (as per the README) as-is on a RHEL/CentOS box as well (@afomm just confirmed it working now).

cevich:
@edsantiago @afomm 

FWIW: ATM, the expected result/output is:

*  No playbook tasks or "hosts" should fail (hosts are all aliases for 'localhost' ATM)
*  adept.py should always exit 0
*  All contexts should end with ``RUNNING HANDLER [exekutir_lock : Release lock on exekutir]``
*  All contexts should have this somewhere in the middle of all the output:

```
TASK [debug] *******************************************************************
ok: [kommandir] => {
    "msg": "This is the playbook running on kommandir for the setup context"
}
```

edsantiago:
I gave the diffs a once-over this morning, for brief familiarity and typo checking, but the rest is looking at the gestalt.

For posterity, I checked for typos via:

    $ for i in kommandir exekutir vmexists teeecho logecho scrapetable headfield; do echo;echo $i;git log -1 -p HEAD^|agrep -i -2 $i|grep -vi $i;done

This is imperfect--it will mask an instance where the given word appears twice in one line, once correctly and once misspelled--but it is sufficient for my purposes.

cevich:
> For posterity, I checked for typos via:

@edsantiago Ohhhh, that's really cool!  I wonder if we could get that into ``.travis.yml`` so it could check this (and future PRs) automaticly.  Thanks!

edsantiago:
No, it can't be automated. Too many false positives that require human attention to dismiss. And how would the keyword list be defined, anyway? I chose it as a sampling of newly-added terms.

cevich:
> No, it can't be automated.

@edsantiago well no, not exactly like that, but we could check for some key words/ and common ytpos that should never appear.

```
TYPOS="komandir executir ecoh roel fixup! squash! FIXME"
for i in $TYPOS; do echo; echo "$i"; git log -1 -p HEAD^ | grep -a -2 "$i"; test "$?" == "1" || exit 1; done
```

.gitignore
cevich: hmmmm, that's a good point about the spellings.  It use to trip me up until seeing/typing them a billion times.  I like how they "pop" because some of the plays are really tricky about the order or which hosts things happen on.  So, having them stand out in that regard isn't a bad thing.

I'm a bit loath to use "slave", because of the possible negative connotations and maybe that it's unimportant or to be disregarded.  The other problem is "commander" is awful similar to the name of an Ansible module, which could cause confusion.  Worse, if you accidentally slip a space into "executor" in just the wrong spot,  will most certainly cause a bad smell to emanate from your shell output 😀 

Though I'm not at all against changing the names to something else.  Do you have any better ideas? 

What about if we could "underline" them instead.  Like "__commander__" and "__executor__"  then they could match the directory names also.  I'm just not sure if ansible would accept that since you can't have underscores in DNS names.

exekutir.xn
```diff
@@ -0,0 +1,140 @@
...
+                echo "Copying ssh keys from $(dirname $ANSIBLE_PRIVATE_KEY_FILE)";
+                cp -a "${ANSIBLE_PRIVATE_KEY_FILE}" "$WORKSPACE/ssh/id_rsa";
+                cp -a "${ANSIBLE_PRIVATE_KEY_FILE}.pub" "$WORKSPACE/ssh/id_rsa.pub";
+            else
+                echo "WARNING: Generating new ssh keys for kommandir in $WORKSPACE/ssh";
+                echo "Do not loose them if you need to access a remote kommandir.";
+                echo "If this is not what you want, set the ANSIBLE_PRIVATE_KEY_FILE";
+                echo "env. var";
+                echo;
+                ssh-keygen -f "$WORKSPACE/ssh/id_rsa" -N "" > /dev/null;
```
edsantiago: This hangs if the id file already exists (with an "overwrite y/n" prompt that is never seen). Solution: see above.

exekutir.xn
```diff
@@ -0,0 +1,140 @@
...
+            echo "Creating $WORKSPACE/ssh";
+            mkdir -p "$WORKSPACE/ssh";
+            ln -sf "$WORKSPACE/ssh" "$WORKSPACE/.ssh"
+            echo "-----";
+            if [ -r "${ANSIBLE_PRIVATE_KEY_FILE}" ];
+            then
+                echo "Copying ssh keys from $(dirname $ANSIBLE_PRIVATE_KEY_FILE)";
+                cp -a "${ANSIBLE_PRIVATE_KEY_FILE}" "$WORKSPACE/ssh/id_rsa";
+                cp -a "${ANSIBLE_PRIVATE_KEY_FILE}.pub" "$WORKSPACE/ssh/id_rsa.pub";
+            else
```
edsantiago: This should be `elif [ ! -e $MY_KEY_FILE ]`; see below

exekutir.xn
```diff
@@ -0,0 +1,140 @@
...
+            # from the $ANSIBLE_PRIVATE_KEY_FILE variable if it was exported.
+            echo "-----";
+            echo "Creating $WORKSPACE/ssh";
+            mkdir -p "$WORKSPACE/ssh";
+            ln -sf "$WORKSPACE/ssh" "$WORKSPACE/.ssh"
+            echo "-----";
+            if [ -r "${ANSIBLE_PRIVATE_KEY_FILE}" ];
+            then
+                echo "Copying ssh keys from $(dirname $ANSIBLE_PRIVATE_KEY_FILE)";
+                cp -a "${ANSIBLE_PRIVATE_KEY_FILE}" "$WORKSPACE/ssh/id_rsa";
```
edsantiago: I'm SUPER-uncomfortable with the string 'rsa' in the name. Maybe `id_kommandir` might be better: it doesn't misleadingly imply rsa. And it should be refactored into an envariable, like MY_KEY_FILE because there's a lot of duplication.

cevich: OIC, you mean it shouldn't copy a possibly non-rsa key to a name with ``rsa`` in it.  Yeah, you're right that could cause confusion.

I can't change the env. var.  ``ANSIBLE_PRIVATE_KEY_FILE`` b/c that's defined by Ansible itself.  Or were you meaning something else?

edsantiago:
[in response to your response re: spellings, exekutir, etc, which I can't find on github any more]

>Though I'm not at all against changing the names to something else.  Do you have any better ideas? 

No, and I think it's probably too late to change them now anyway. I just never spoke up about kommandir, and now there's another weird one, and I would prefer not to see it become a convention.

exekutir.xn
edsantiago: non-rsa (eg ed25519) with rsa: exactly.

env var: no, I mean, assign "$WORKSPACE/ssh/id_whatever" into a var so it doesn't have to be repeated.

cevich:
> No, and I think it's probably too late

@edsantiago why is it too late?  AFAIK, nothing outside of this repo. references those names in any way that can't also be updated (likely only comments).  I'm trying to be sensitive about your comment, so if you really feel strongly that they should be changed, I'll work to make it happen.

exekutir.xn
cevich: ahh, yep, good catch.  fixing it...

exekutir.xn
cevich: Ooohhhh, good idea.

cevich: okay, fixed.

exekutir.xn
cevich: fixed.

exekutir.xn
cevich: fixed, and removed the ``> /dev/null``, replaced with ``-q`` instead.

cevich:
@edsantiago BTW: I've added something like your 'typo check' to ``.travis.yml`` along with making it actually run the playbook (CI -CI, FTW!).

```
    # Check for some typos / git commit problems
    - for i in $TYPOS; do echo; echo "$i"; git log -1 -p HEAD^ | grep -a -2 "$i"; test "$?" == "1" || exit 1; done
    # Test building docs
    - cd docs && make html; cd ..
    - unit2 -vfc
    - mkdir ../travis_ci_adept_workspace
    # Relative path
    - ./adept.py setup ../travis_ci_adept_workspace exekutir.xn -v
    # absolute path
    - ./adept.py run ${PWD}/../travis_ci_adept_workspace exekutir.xn
    # current path
    - cd ../travis_ci_adept_workspace && ./adept.py cleanup . exekutir.xn
```

@afomm and @edsantiago how are all the comments I'm sticking everywhere, helpful?  Need more? Less? Good 'nuf?

Thanks for the help guy!

cevich:
> Great! I think a demo/conference would be great. 

Okay! Let's do it.  We can call it a "tour" if you like b/c you've already demoed it to yourselfs 😁

cevich:
**N/B**: Travis identified an ansible-1.9 compatibility problem.  I committed a FIXME into one of the plays that will cause everything to break until it's actually fixed 😁

cevich:
There, travis is happy now.  I had to bump the minimum up to Ansible 2.1.  Tried really hard to make 1.9 work, but the ``add_host`` module simply behaves in an odd way that would force us to use dynamic inventory (which I'm trying to avoid for simplicity reasons).

__exekutir__/roles/exekutir_workspace_setup/tasks/main.yml
```diff
@@ -0,0 +1,73 @@
...
+    src: "{{ kommandir_workspace }}/ssh"
+    dest: "{{ kommandir_workspace }}/.ssh"
+    state: link
+  when: kommandir_workspace_dirty
+
+- name: Future kommandirs and peons protected by new ssh key if kommandir_workspace_dirty
+  command: ssh-keygen -f "{{ kommandir_workspace }}/ssh/kommandir_key" -N ""
+  args:
+    chdir: "{{ kommandir_workspace }}/ssh/"
+    creates: "{{ kommandir_workspace }}/ssh/id_rsa"
```
edsantiago: Should be id_kommandir

cevich: ahh, good catch, yes.  Actually it should be ``kommandir_key``

cevich:
once again for the second time...

cevich:
Only a rebase + squash, nothing to see here.

cevich:
uggg. DNR...attempting to fix conflicts...

cevich:
Okay, conflicts fixed.

cevich:
rebased from master, added some simple fixes, --force pushed.

cevich:
I've killed all the old comments.  Working on the last little bits needed to get a temporary internal-CI job running against PRs made to this branch.  Then can switch over to the standard PR-workflow for contributions to this branch while I jump neck-deep into documenting.

PR #103, opened 2017-02-10
Rename/Change yumrepos gpgcheck input arg. name.

cevich:
No idea why 'exclude' was used, probably a copy-paste typo on my part.
Also defaulting to 'False' (i.e. --nogpgcheck) isn't the correct
behavior.  It's important to spot instances where packages are unsigned,
and they should be.  This change will cause those cases to (correctly)
fail.

For cases where development/test repos. are being used, their
list item args will need to be updated with 'gpgcheck: False'
to get the correct (previous) behavior.  Assuming that devel/test
repos don't generally sign packages.

### DIFF
--- a/roles/yumrepos/defaults/main.yml
+++ b/roles/yumrepos/defaults/main.yml
@@ -1,6 +1,25 @@
 ---
 
+# There are situations where default repositories are broken.
+# Setting this true causes ALL subscription-manager supplied repos
+# to be disabled.
 disable_all_rh_repos: False
+
+# This is the opposite of ``disable_all_rh_repos`` (above).  It's
+# a list of subscription-manager supplied repos to explicitly enable.
 enable_rh_repos: []
-# empty-list is (unset) sentinel value
+
+# When neither of the above meet testing requirements, or additional
+# local repositories should be added, this specifies them.  Each item
+# in the list is a dictionary of arguments to the standard 
+# ``yum_repository`` Ansible module.  e.g.
+#
+# yum_repos:
+#     - name: "My special Repo"
+#       baseurl: "https://my.special.repo.example.com/"
+#       gpgcheck: False
+#       includepkgs: rocket2moon
+#     - name: "Other special repo"
+#       baseurl: "https://you.get.the.idea"
+#
 yum_repos: []
--- a/roles/yumrepos/tasks/main.yml
+++ b/roles/yumrepos/tasks/main.yml
@@ -15,7 +15,7 @@
     name: "{{ item.name | mandatory }}"
     baseurl: "{{ item.baseurl }}"
     description: "Ansible added {{ item.name | mandatory }} repo"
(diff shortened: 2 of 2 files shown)


edsantiago:
Sounds reasonable at first, but this file seems no longer to exist in the simplify branch. And even if I look on master, I see references to `yum_repos` but no definition anywhere or even examples of what it might contain; this makes it hard to verify that `gpgcheck` (and `baseurl` and actually everything) are valid fields.

But back to my original comment: if this file is gone in the simplify branch, why spend time on it?

cevich:
> But back to my original comment: if this file is gone in the simplify branch, why spend time on it?

This role _should_ be included in simplify, maybe buried under 'old_roles'.  If not, it will get picked up eventually, it's needed for setting up the peons.  In any case, simplify is also unknown-distance out from being complete/ready.  What we have now, as ugly as it is, still needs to help find problems.  This one slipped passed me.

edsantiago:
Yes, it's in old_adept_roles -- I am treating the word "old" as meaning "do not look in here, this is just stuff that is about to be deleted but we need to make sure first that everything works".

Would it be possible to add a sample `yum_repos` somewhere in here so a human can look at what it's supposed to look like?

cevich:
> Yes, it's in old_adept_roles -- I am treating the word "old" as meaning "do not look in here, this is just stuff that is about to be deleted but we need to make sure first that everything works".

Ohhhh, okay, that's an easy thing to fix.  Lets  rename ``old_adept_roles`` -> ``fixme_migrade_roles``

(or something to that effect)

> Would it be possible to add a sample yum_repos somewhere in here so a human can look at what it's supposed to look like?

Yep, also good feedback.  Normally I'd stick this kind of thing into ``defaults`` however, in this case the needed default is "do nothing".  I can certainly stick it into a comment though, no prob.

cevich:
``--force`` pushed w/ documentation in ``defaults/main.yml`` comments.

roles/yumrepos/defaults/main.yml
```diff
@@ -1,6 +1,25 @@
...
+# in the list is a dictionary of arguments to the standard 
+# ``yum_repository`` Ansible module.  e.g.
+#
+# yum_repos:
+#     - name: "My special Repo"
+#       baseurl: "https://my.special.repo.example.com/"
+#       gpgcheck: False
+#       includepkgs: rocket2moon
+#     - name: "Other special repo"
+#       baseurl: "https://you.get.the.idea
```
edsantiago: Not strictly necessary, but a closing quote would be welcome

cevich: woops.  thanks.

cevich:
fixed and --force pushed

cevich:
great thanks, I appreciate the expert application of your thoroughly calibrated vision-orbs 😀

PR #105, opened 2017-02-10
fixup! Rename old_adept_roles -> fixme_migrate_roles

cevich:
Clarifies that they're not to be deleted, but will eventually need
migrating to the new job.xn scheme (when it's fully written).


cevich:
***N/B:***  This is against the ``simplify`` branch.

PR #106, opened 2017-02-13
Atomic requires use of a rsyslog system container

cevich:
Long ago, bits were stripped from Atomic to make it smaller.  That
included removing rsyslog (needed to flush from journal to messages).
Though I have no idea why the syslog subtest has been passing on CentOS
Atomic, it's not a setup that we need to test.  Exclude the syslog
subtest from testing on Atomic hosts, only run it on full-OS systems.

### DIFF
--- a/roles/autotested/templates/control.ini.j2
+++ b/roles/autotested/templates/control.ini.j2
@@ -18,8 +18,10 @@ exclude = example,subexample,pretest_example,intratest_example,
           posttest_example,
           docker_cli/run_volumes,
 {# Atomic host does not have git #}
+{# Atomic host has trimmed down logging #}
           {% if is_atomic %}
           docker_cli/build/git_path,
+          docker_cli/syslog
 {% endif %}
 {% if not is_enterprise %}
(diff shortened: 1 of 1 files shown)


cevich:
Hmmm, not sure about this.  Just checked and for sure CentOS Atomic has a /var/log/messages.

cevich:
damnit...

edsantiago:
Did you mean for this (2nd commit) to be a separate PR?

cevich:
> Did you mean for this (2nd commit) to be a separate PR?

ya, totally screwed it up.  was adding comments to the defaults file (just how you like) and pushed the wrong branch 😖

cevich:
going to sit on this PR a bit, I may just close it.

edsantiago:
I would agree with closing. If syslog problems persist after my docker-autotest syslog fix, then reinvestigate. It is not worth spending time on any of this until docker-autotest syslog test is fixed.

cevich:
If it ain't broke, don't fix it.

PR #107, opened 2017-02-13
Fix: timeouts during install/upgrade

cevich:
Ansible docs state "operations that take longer than the ssh
timeout will fail".  This is certainly more likely to happen with large
package installs/atomic updates depending on networking conditions.
Reduce dependency on low-latency networking by monitoring the
install/update via polling.

### DIFF
--- a/roles/atomic_upgrade/defaults/main.yml
+++ b/roles/atomic_upgrade/defaults/main.yml
@@ -1,3 +1,6 @@
 ---
 
 rpms_updated: False
+
+# Max time to poll for update/install results on remote host.
+async_poll_timeout: 300
--- a/roles/atomic_upgrade/tasks/main.yml
+++ b/roles/atomic_upgrade/tasks/main.yml
@@ -50,6 +50,9 @@
       async: '{{ async_poll_timeout }}'
       poll: 10
 
+    - debug:
+        var: 'atomic_host_upgrade'
+
     - assert:
         that:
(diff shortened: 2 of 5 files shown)


cevich:
there, --force pushed (to the right branch this time):  Added documentation comment to role defaults.

cevich:
Hmmm, this isn't right:

```

13:27:01 TASK [atomic_upgrade : Atomic host is updated to latest tree] ******************

13:29:24 fatal: [centosatomic-7-2-ADEPT-ci-202-120609d76e7d]: FAILED! => {"failed": true, "msg": "The conditional check 'atomic_host_upgrade.rc in [0,77]' failed. The error was: error while evaluating conditional (atomic_host_upgrade.rc in [0,77]): 'dict object' has no attribute 'rc'"}

13:29:24 fatal: [centosatomic-latest-7-2-ADEPT-ci-202-9bea24709e80]: FAILED! => {"failed": true, "msg": "The conditional check 'atomic_host_upgrade.rc in [0,77]' failed. The error was: error while evaluating conditional (atomic_host_upgrade.rc in [0,77]): 'dict object' has no attribute 'rc'"}
```

cevich:
damn, that wasn't it.

cevich:
ahh, leftover ``retries: 3`` forgot to remove.  Still only suspecting that's causing the ansible failure (i.e. a command/shell task with a register always has an ``rc`` attribute.)

PR #109, opened 2017-02-15
fixup!  Rename jobs -> __jobs__

cevich:
This directory is really important, make sure it stands out among all
the others.  Also, since it contains playbooks and inventory files,
give it the same look as other directories containing the same.


cevich:
@edsantiago in your feedback good/bad/indifferent.  IMHO it's a very minor improvement.  Another thought I had was to just drop all the ``__dir__`` names and just use ``dir``.  i.e. I'm not sure the underlining helps anything.

edsantiago:
Gahhh... I can't review this, sorry. I haven't yet looked at the jobs directory and was hoping to postpone that. IMO the underlining would be useful if there was a huge amount of top-level clutter, but there isn't.

cevich:
> useful if there was a huge amount of top-level clutter, but there isn't.

Aye, that was my thinking as well.  I think I'll kill them all then.

PR #110, opened 2017-02-15
fixup! Simplify fix fatal exit

cevich:
The first commit tries to explain what's fixed here (it's an ansible-ism).  The two other commits are just supporting changes and related uninttest updates.

When any task fails, ansible skips all other tasks for that host, but
continues with tasks for other hosts.  Because pre-job.xn setup tasks
are mandatory, change the default behavior of ansible to exit
immediatly when any task fails.  The play argument
``any_errors_fatal: true`` accomplishes this.

The exception to the above is during the cleanup transition.  That
context must always complete (successful or not).  In other words,
whether or not kommandir re/discovery is successful, there are still
additional tasks to perform on the exekutir.  Allow those cases
by NOT setting ``any_errors_fatal`` in cleanup_before_job.yml

### DIFF
--- a/__exekutir__/cleanup_before_job.yml
+++ b/__exekutir__/cleanup_before_job.yml
@@ -1,3 +1,41 @@
 ---
 
-- include: setup_before_job.yml
+- hosts: exekutir
+  gather_facts: False
+  pre_tasks:
+    - name: Gather exekutir facts exclusive of other inventory hosts
+      setup:
+  roles:
+    - common
+    - ansible_versioned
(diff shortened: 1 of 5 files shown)


cevich:
@edsantiago if you dunna wanna try and understand this, it's NP.  unittests pass, and documentation (should) still match behavior (if not, it's a bug).  Crap...double-checking docs...

edsantiago:
I'll pass, thanks. Too much churn again.

PR #111, opened 2017-02-15
fixup! Can't check out tag w/o enough depth

cevich:
By default the fetch depth is set at 2.  This is not deep enough to
locate the tag in the case of python-bugzilla.  Add an argument to the
configured list so the depth is left "None" - meaning fetch
everything.


cevich:
@afomm said he would make a PR that fixes this properly.

PR #114, opened 2017-02-16
Modifying git depth

afomm:
Since git module_arg depth:2 fails for python-bugzilla repo, omitting it for this repo and setting it individually for the other repos. Please review.

### DIFF
--- a/__exekutir__/inventory/host_vars/kommandir.yml
+++ b/__exekutir__/inventory/host_vars/kommandir.yml
@@ -87,4 +87,6 @@ git_cache_args:
     - repo: "https://github.com/python-bugzilla/python-bugzilla.git"
       recursive: False
       cachepath: "python-bugzilla"
+      # Need to fetch everything to find specific tagged version
+      depth:
       version: "v1.2.2"
--- a/__exekutir__/inventory/host_vars/kommandir.yml
+++ b/__exekutir__/inventory/host_vars/kommandir.yml
@@ -87,6 +87,6 @@ git_cache_args:
     - repo: "https://github.com/python-bugzilla/python-bugzilla.git"
       recursive: False
       cachepath: "python-bugzilla"
-      # Need to fetch everything to find specific tagged version
-      depth:
+      # Do not include a 'depth' argument, git needs the full history to be 
+      # able to locate specific commits/branches/tags referenced by 'version'
       version: "v1.2.2"
--- a/__exekutir__/roles/kommandir_workspace_setup/tasks/main.yml
+++ b/__exekutir__/roles/kommandir_workspace_setup/tasks/main.yml
@@ -123,8 +123,9 @@
   git:
     repo: "{{ item.repo }}"
     dest: "{{ kommandir_workspace }}/cache/{{ item.cachepath }}"
-    # Shallow clone is much faster, two commits will cover a merge-commit
-    depth: "{{ item.depth | default(2) }}"
+    # Shallow clone is much faster when specific branch/tag/version is _not_ used.
+    # Otherwise, if a version is specified, then the full history-depth is likely required to find it
+    depth: "{{ omit if item.version is defined else item.depth | default(2) }}"
     # HEAD is the default if not specified
     version: "{{ item.version | default(omit) }}"
     recursive: "{{ item.recursive | default(False) }}"


cevich:
@afomm looks great.  Can you just ``--amend`` and pre-pend a ``fixup!`` at the front of your commit message?  Then I can merge this and close my PR.  Thanks!

__exekutir__/roles/kommandir_workspace_setup/tasks/main.yml
```diff
@@ -123,8 +123,9 @@
   git:
     repo: "{{ item.repo }}"
     dest: "{{ kommandir_workspace }}/cache/{{ item.cachepath }}"
-    # Shallow clone is much faster, two commits will cover a merge-commit
-    depth: "{{ item.depth | default(2) }}"
+    # Shallow clone is much faster, two commits will cover a merge-commit and will be set per repo.
+    # default(omit) to workaround failure with python-bugzilla repo
+    depth: "{{ item.depth | default(omit) }}"
```
edsantiago: This makes me uncomfortable. It adds too much dependency on kommandir.yml, and in particular, requiring that the maintainer magically know to add 'depth: 2' in some places but not if there's a specific tag.

Since the reason for this change is (IIUC): "do a shallow copy if we can, but we can't do that for tagged versions", how about:

    depth: "{{ omit if item.version is defined else item.depth | default(2) }}"   **untested**

I haven't tested this, and am not sure how to do so, and am particularly concerned about the `|` because I don't know how it combines with `omit`. If that fails, maybe this could work:

    depth: "{{ item.depth | default(omit if item.version is defined else 2) }}"

The point is, this is a one-time edit that will do the right thing and never cause problems when, in six months, someone adds a new repo with a version and a depth.

__exekutir__/inventory/host_vars/kommandir.yml
```diff
@@ -81,10 +81,13 @@ git_cache_args:
...
       cachepath: 'autotest'
+      depth: 2
     - repo: "https://github.com/autotest/autotest-docker.git"
       recursive: True
       cachepath: 'autotest-docker'
+      depth: 2
     - repo: "https://github.com/python-bugzilla/python-bugzilla.git"
       recursive: False
       cachepath: "python-bugzilla"
+      # Need to fetch everything to find specific tagged version
```
edsantiago: I realize that this comment is supposed to be interpreted as "this is why you're not seeing a `depth` parameter here", but a future reviewer will think that the comment applies to the `version` line... and that makes no sense. 

If my one-line-conditional suggestion below works out, then you can just remove this comment and life is good.

If my one-line-conditional suggestion does not work, and we have to live with this code, then this comment should be updated to say something like "Do not include a depth parameter here! [...this is why...]"

__exekutir__/roles/kommandir_workspace_setup/tasks/main.yml
cevich: Interesting idea, and your motivations are good.  The only way I know of is by hacking in jinja2 templates.  Though I wonder if what you have would work (I've not tried it).  Something like this would do what you're suggesting, but as you can see, it's a bit fugly (why I like to avoid it):

```yaml
...
- name: Kommandir's cache content checked out from git when git_cache_args non-empty, result is false, and both kommandir_workspace_dirty flags set
  git:
    repo: "{{ item.repo }}"
    dest: "{{ kommandir_workspace }}/cache/{{ item.cachepath }}"
    # Shallow clone is much faster, two commits will cover a merge-commit
    depth: |
        {%- if item.version is defined and item.version not in [None,'',[],{}] -%}\
            {{ omit }}\
        {%- else -%}\
            {{ item.depth | default(2) }}\
        {%- endif -%}
    # HEAD is the default if not specified
    version: "{{ item.version | default(omit) }}"
    recursive: "{{ item.recursive | default(False) }}"
  when: kommandir_workspace_dirty and hostvars.exekutir.kommandir_workspace_dirty and
        not result and git_cache_args is defined
  with_items: '{{ git_cache_args }}'
  ...
```

My experience with Ansible is, the more you try and stuff it into the "programming language" box, the more it will cause general misery / frustration.  Another way is to write the above into an actual jinja2 template file, and then read it in using ``{{ lookup('template','path/to/file') }}``.  The down-side is it means the future engineer has to go hunting for that file to understand what the task does.

I'm totally open to fixing this in the spirit of what you suggest.  Just ignorant of a good way to do it.  Also, to be sure, there are probably many more areas where a similar maintainability-advice could be applied.  I'd like to find/fix all of them if we can, just not sure of what's the best/better way.  Can I open an issue to remind me/we/us to play with it, and reference your example/suggestion?

__exekutir__/inventory/host_vars/kommandir.yml
cevich: This is my fault, I totally forgot to update that comment.  How about:

```# Do not include a 'depth' argument, git needs the full history to be able to locate specific commits/branches/tags referenced by 'version'```

File: __exekutir__/roles/kommandir_workspace_setup/tasks/main.yml (hunk 1 above)
Comment (cevich) [inline __exekutir__/roles/kommandir_workspace_setup/tasks/main.yml]: @edsantiago docs ref: http://jinja.pocoo.org/docs/2.9/templates/#list-of-control-structures
(I keep a copy under my pillow)

Comment (edsantiago) [inline __exekutir__/roles/kommandir_workspace_setup/tasks/main.yml]: Can we just try my suggestion, then if it doesn't work, go into further discussion?

File: __exekutir__/roles/kommandir_workspace_setup/tasks/main.yml [hunk 3]
```diff
@@ -123,8 +123,9 @@
   git:
     repo: "{{ item.repo }}"
     dest: "{{ kommandir_workspace }}/cache/{{ item.cachepath }}"
-    # Shallow clone is much faster, two commits will cover a merge-commit
-    depth: "{{ item.depth | default(2) }}"
+    # Shallow clone is much faster, two commits will cover a merge-commit and will be set per repo.
+    # default(omit) to workaround failure with python-bugzilla repo
```
cevich: This comment about specific failure doesn't belong in general-purpose code.  How about:

```
# Shallow clone is much faster when specific branch/tag/version is _not_ used.
# Otherwise, if a version is specified, then the full history-depth is likely required to find it
```

__exekutir__/roles/kommandir_workspace_setup/tasks/main.yml
cevich: Hmmm, maybe it could work.

```
$ ansible -i /dev/null -c local localhost -m debug -a 'msg="{{ omit if true else notexist | default(2) }}"'
```

Outputs "msg: Hello World" (which is the 'omit' behavior)

cevich: ahh, I C now, and here's the docs for including ``if``` in expressions:  http://jinja.pocoo.org/docs/2.9/templates/#if-expression

cevich: @afomm give Ed's suggestion (the first one) a try.

``depth: "{{ omit if item.version is defined else item.depth | default(2) }}" ``

Neat!  I was worried I might not learn something new today, thanks Ed!

afomm: It's working for me also. I'll update it. Thanks Ed!

cevich: Great!  A team effort, and a schooling by Ed 😀

cevich:
Thanks @afomm and @edsantiago I really appreciate you're taking an active hand in this stuff.

__exekutir__/inventory/host_vars/kommandir.yml
```diff
@@ -87,4 +87,6 @@ git_cache_args:
     - repo: "https://github.com/python-bugzilla/python-bugzilla.git"
       recursive: False
       cachepath: "python-bugzilla"
+      # Do not include a 'depth' argument, git needs the full history to be 
+      # able to locate specific commits/branches/tags referenced by 'version'
```
edsantiago: This comment isn't really needed any more. It was only needed in the previous revision, when other blocks had a `depth` entry.

afomm: @edsantiago I was thinking of moving it up right under "git_cache_args:" but left it. You think removing it better?

edsantiago: @afomm if the `depth` option were actually documented anywhere, then yes, it would make sense to extend that documentation with this caveat. But (1) it's not documented; and (2) even if someone _does_ include a depth argument, the new code will gracefully ignore it, so there isn't really a need for the warning. I think removing it is the best option.

afomm: @edsantiago ok removing it. Unless @cevich has any objection.

afomm: @edsantiago PR116 - https://github.com/cevich/autotest-docker-enabled-product-testing/pull/116.

cevich: Documentation for that option is here: http://docs.ansible.com/ansible/git_module.html#options

edsantiago: No documentation _in this module_. There is no explanation that `git_cache_args` will be passed to the `git` module, no explanation of the arguments, no explanation of anything. Whether one would be useful or not is beyond the scope of this discussion. My point is: if there were documentation in this file, and that documentation included the `depth` option, then a "this will be ignored if you also give a version" comment would be helpful. Absent documentation, the comment is unhelpful.

cevich: > No documentation in this module.

Ahh, yes.  Do you think it would be more useful there (in a comment) or in the proper Sphinx-based (pretty) docs?  (I was thinking of starting a "variables reference" section)

edsantiago: Comments are always most helpful when they're close to the relevant code. But again, I don't think this is worth spending time on right now.

PR #115, opened 2017-02-17
Add openstack kommandir provisioning script

cevich:
Work-in-progress (it's a mess).  Releasing early for review/feedback (yes I know it needs functions).  Incomplete and minimally tested at this point (copy-paste into python CLI).  Only supports kommandir discovery (not creation or removal yet).  Requires ``clouds.yml`` to exist in ``$WORKSPACE`` (as in, you put it there).  Though supporting ``OS_*`` env. vars. can be done as well.  I'm not sure if I want to continue using the REST API (portable, less library version sensitive, very fast) or just shell-out to the ``openstack`` CLI (ease of implementation, but possibly more version sensitive)

### DIFF
--- a/kommandir/bin/adept_openstack.py
+++ b/kommandir/bin/adept_openstack.py
@@ -1,0 +1,711 @@
+#!/usr/bin/env python2
+
+"""
+ADEPT VM provisioning / cleanup script for the openstack ansible group.
+
+Assumed to be running under an exclusive lock to prevent TOCTOU race
+and/or clashes with existing named VMs.
+
+Requires:
+*  RHEL/CentOS/Fedora host w/ RPMs:
+    * redhat-rpm-config (base)
+    * python-virtualenv or python2-virtualenv (EPEL)
+* Openstack credentials as per:
+    https://docs.openstack.org/developer/os-client-config/
+* Execution under adept.py exekutir.xn transition file or unittests
+"""
+
+import sys
+import os
+import os.path
+import logging
+import subprocess
+import time
+from base64 import b64encode
+
+# Placeholder, will be set by unittests or under if __name__ == '__main__'
+os_client_config = ValueError  # pylint: disable=C0103
+
+# Lock-down versions of os-client-config and all dependencies for stability
+PIP_REQUIREMENTS = [
+    'os-client-config==1.21.1',
+    'appdirs==1.4.2',
+    'iso8601==0.1.11',
+    'keystoneauth1==2.18.0',
+    'os-client-config==1.21.1',
+    'pbr==1.10.0',
+    'positional==1.1.1',
+    'PyYAML==3.12',
+    'requests==2.13.0',
+    'requestsexceptions==1.1.3',
+    'six==1.10.0',
+    'stevedore==1.20.0',
+    'wrapt==1.10.8']
+# No C/C++ compiler is available in this virtualenv
+PIP_ONLY_BINARY = [':all:']
+# These must be compiled, but don't require C/C++
+PIP_NO_BINARY = ['wrapt', 'PyYAML', 'positional']
+
+IMAGE = 'CentOS-Cloud-7'
+FLAVOR = 'm1.medium'
+
+# Exit code to return when --help output is displayed (for unitesting)
+HELP_EXIT_CODE = 127
+
+# Operation is discovered by symlink name used to execute script
+DISCOVER_CREATE_NAME = 'openstack_discover_create.py'
+DESTROY_NAME = 'openstack_destroy.py'
+
+OUTPUT_FORMAT = """---
+inventory_hostname: {name}
+ansible_host: {floating_ip}
+ansible_ssh_host: {{{{ ansible_host }}}}
+"""
+
+class OpenstackREST(object):
+    """
+    State full cache of Openstack REST API interactions.
+
+    :docs: https://developer.openstack.org/#api
+    """
+
+    # The singleton
+    _self = None
+
+    # Cache of current and previous response instances and json() return.
+    response_json = None
+    response_obj = None
+    # Useful for debugging purposes
+    previous_responses = None
+
+    # Current session object
+    service_sessions = None
+
+    def __new__(cls, service_sessions=None):
+        if cls._self is None:  # Initialize singleton
+            if service_sessions is None:
+                raise ValueError("service_sessions must be passed the first time")
+            cls._self = super(OpenstackREST, cls).__new__(cls)
+            cls.service_sessions = service_sessions
+            cls.previous_responses = []
+        return cls._self  # return singleton
+
+    def raise_if(self, true_condition, xception, msg):
+        """
+        If true_condition, throw xception with additional msg.
+        """
+        if not true_condition:
+            return None
+
+        if self.previous_responses:
+            responses = self.previous_responses + [self.response_obj]
+        elif self.response_obj:
+            responses = [self.response_obj]
+        else:
+            responses = []
+
+        logging.debug("Response History:")
+        for response in responses:
+            logging.debug('  (%s) %s:', response.request.method, response.request.url)
+            try:
+                logging.debug('    %s', response.json())
+            except ValueError:
+                pass
+            logging.debug('')
+
+        if callable(xception):  # Exception not previously raised
+            xcept_class = xception
+            xcept_value = xception(msg)
+            xcept_traceback = None   # A stack trace would be nice here
+            raise xcept_class, xcept_value, xcept_traceback
+        else:  # Exception previously raised, re-raise it.
+            raise
+
+    def service_request(self, service, uri, unwrap=None, method='get', post_json=None):
+        """
+        Make a REST API call to uri, return optionally unwrapped json instance.
+
+        :param service: Name of service to request uri from
+        :param uri: service URI for get, post, delete operation
+        :param unwrap: Optional, unwrap object name from response
+        :param method: Optional, http method to use, 'get', 'post', 'delete'.
+        :param post_json: Optional, json instance to send with post method
+        :raise ValueError: If request was unsuccessful
+        :raise KeyError: If unwrap key does not exist in response_json
+        :returns: json instance
+        """
+        session = self.service_sessions[service]
+        if self.response_obj is not None:
+            self.previous_responses.append(self.response_obj)
+        if method == 'get':
+            self.response_obj = session.get(uri)
+        elif method == 'post':
+            self.response_obj = session.post(uri, json=post_json)
+        elif method == 'delete':
+            self.response_obj = session.delete(uri)
+        else:
+            self.raise_if(True,
+                          ValueError,
+                          "Unknown method %s" % method)
+
+        code = int(self.response_obj.status_code)
+        self.raise_if(code not in [200, 201, 202, 204],
+                      ValueError, "Failed: %s request to %s: %s" % (method, uri, code))
+
+        try:
+            self.response_json = self.response_obj.json()
+        except ValueError:  # Not every request has a JSON response
+            self.response_json = None
+
+        # All responses encode the object under it's name.
+        if unwrap:
+            self.raise_if(unwrap not in self.response_json,
+                          KeyError, "No %s in json: %s"
+                          % (unwrap, self.response_json))
+            self.response_json = self.response_json[unwrap]
+            return self.response_json
+        else:  # return it as-is
+            return self.response_json
+
+    def compute_request(self, uri, unwrap=None, method='get', post_json=None):
+        """
+        Short-hand for ``service_request('compute', uri, unwrap, method, post_json)``
+        """
+        return self.service_request('compute', uri, unwrap, method, post_json)
+
+    def child_search(self, key, value=None, alt_list=None):
+        """
+        Search cached, or alt_list for object with key, or key set to value.
+
+        :param key: String, the key to search for, return list of values.
+        :param value: Optional, required value for key, first matching item.
+        :param alt_list: Optional, search this instance instead of cache
+        :raises TypeError: If self.response_json or alt_json are None
+        :raises ValueError: If self.response_json or alt_json are empty
+        :raises IndexError: If no item has key with value
+        """
+        if alt_list is not None:
+            search_list = alt_list
+        else:
+            search_list = self.response_json
+
+        self.raise_if(search_list is None,
+                      TypeError, "No requests have been made/cached")
+
+        if value:
+            found = [child for child in search_list
+                     if child.get(key) == value]
+            try:
+                return found[0]
+            except IndexError, xcept:
+                self.raise_if(True,
+                              xcept,
+                              'Could not find key %s with value %s in %s'
+                              % (key, value, search_list))
+        else:
+            found = [child[key] for child in search_list
+                     if key in child]
+            self.raise_if(not found,
+                          IndexError,
+                          'Could not find key %s in %s'
+                          % (key, search_list))
+            return found
+
+    def server_list(self):
+        """
+        Cache and return list of server names or empty list
+
+        :returns: list of strings
+        """
+        self.compute_request('/servers', 'servers')
+        try:
+            return self.child_search('name')
+        except IndexError:
+            return []
+
+    def server(self, name=None, uuid=None):
+        """
+        Cache and return details about server name or uuid
+
+        :param name: Optional, exclusive of uuid, name of server
+        :param uuid: Optional, exclusive of name, ID of server
+        :returns: dictionary of server details
+        """
+        self.raise_if(not name and not uuid,
+                      ValueError,
+                      "Must provide either name or uuid")
+        if name:
+            self.compute_request('/servers', 'servers')
+            server_details = self.child_search(key='name', value=name)
+            uri = '/servers/%s' % server_details['id']
+        elif uuid:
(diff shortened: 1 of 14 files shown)


__kommandir__/bin/adept_openstack.py
```diff
@@ -0,0 +1,100 @@
...
+matched_server = [s for s in servers_json
+                  if s['name'] == NAME]
+
+
+if len(matched_server) > 1:
+    raise ValueError("Found more than one server named %s" % NAME)
+elif len(matched_server) == 1:
+    matched_server = matched_server[0]
+    response = session.get('/servers/%s' % matched_server['id'])
+    matched_server = response.json()['server']
```
edsantiago: General practice suggestion: `matched_server` is three different types over the course of just a few lines. That signals a good time to rethink variable names or perhaps to refactor into functions.

__kommandir__/bin/adept_openstack.py
```diff
@@ -0,0 +1,100 @@
...
+    MODE='discover'
```
edsantiago: I thought the rule was that CAPITAL LETTERS were for constants, and constants didn't change?

__kommandir__/bin/adept_openstack.py
```diff
@@ -0,0 +1,100 @@
...
+else:
+    MODE='create'
+    matched_server = None
+
+
+if MODE == 'discover':
```
edsantiago: You can probably get rid of this use of discover/create by simply testing `if matched_server:`

__kommandir__/bin/adept_openstack.py
```diff
@@ -0,0 +1,100 @@
...
+                # Prevent creating the default, generic user
+                "users:\n"
+                "   - name: root\n"
+                "     primary-group: root\n"
+                "     homdir: /root\n"
+                "     system: true\n")
+
+    # ref: https://developer.openstack.org/api-ref/compute/?expanded=create-server-detail#create-server
+else:  # mode == destroy
+    pass
```
edsantiago: I'm fond of the python `if __main__` convention; it makes it easy to write unit tests.

cevich:
> not a real review, just suggestions for before this gets too big

Exactly what I was looking for, thanks 😀

__kommandir__/bin/adept_openstack.py
cevich: The thought had crossed my mind a few times, thanks for underlining it 👍

__kommandir__/bin/adept_openstack.py
cevich: Aye, you caught me.   I'm not very fond of using ``MODE`` anyway, and the same-file w/ symlinks maybe overrated.  Okay, I'll add this to the chopping-block as well.  Thanks.

__kommandir__/bin/adept_openstack.py
cevich: Hehe, that's how it use to be. I think I may just axe ``MODE`` all together.

__kommandir__/bin/adept_openstack.py
cevich: As am I, and yes, unittests, good suggestion, totally forgot about those.

Yes, I plan on writing proper functions/classes soonish, just wanting to get something dirty/working first so I can develop an eye of how/where to split it up.

Thanks for the feedback so far, I'll keep tinkering.

cevich:
Changes:
*  Wrote two whole new functions 😁 
*  Added 11-billion FIXME's
*  Got server creation working (no floating ip yet)

```bash
$ . ~/openrc.sh
$ __kommandir__/bin/openstack_discover_create.py foobar ~/.ssh/id_rsa.pub
State: building Power: 0
State: building Power: 0
State: building Power: 0
State: building Power: 0
State: building Power: 0
State: building Power: 0
State: building Power: 0
State: building Power: 0
State: active Power: 1
```

cevich:
Made code more modular, added some minimal unittests.

cevich:
[(2) (CN) ADEPT Slave Simplification](https://trello.com/c/CevYDUei/279-2-cn-adept-slave-simplification)

cevich:
[(3) (CN) ADEPT Slave VM Provisioning Simplification](https://trello.com/c/tQv1rE9A/420-3-cn-adept-slave-vm-provisioning-simplification)

cevich:
@edsantiago Okay, finally I think this is "good enough".  The script is able to provision a kommandir VM and it has a full and nearly complete set of unittests.

Mind checking if you see any glaring/major problems or gaffs?

I'll be working on this script more next sprint, but I'd love to incorporate your feedback.  The next step is to allow it to accept options for an image name, along with whether or not to create & attach a volume.

**note:**  The diff is near impossible to look at, so just check out the file directly:  https://github.com/cevich/autotest-docker-enabled-product-testing/blob/b949436af103684c94b24e56f4a7d6a01744251a/kommandir/bin/adept_openstack.py

The unittests are here (and all pass):  https://github.com/cevich/autotest-docker-enabled-product-testing/blob/b949436af103684c94b24e56f4a7d6a01744251a/test_openstack.py

Thanks!

edsantiago:
Quick comment: I'm reviewing, looks good so far, but note that it does not pass selftest on f25: pylint barfs. 

One fix is easy: in `test_openstack.py` move "import simplejson" two lines down, to just below urlunparse. The other problem, I can't figure out:

    adept_openstack.py:122: error (E0704, misplaced-bare-raise, OpenstackREST.raise_if) The raise statement is not inside an except clause
FAIL

cevich:
@edsantiago Okay, got your "move it two lines down" implemented (other PR).

cevich:
Ooohhh, ``adept_openstack.py:122: error (E0704, misplaced-bare-raise, OpenstackREST.raise_if) The raise statement is not inside an except clause`` That's coming from pylint eh?  I know why that's happening:  Because Pylint!  (The code makes sure there was a previous exception raised, but pylint isn't that smart).

kommandir/bin/adept_openstack.py
```diff
@@ -0,0 +1,720 @@
...
+            return False
+
+    def action(self):
+        """Collect details about VM status, return server's UUID"""
+        self.os_rest.server(name=self.name)
+        self.vm_state = self.os_rest.response_json['OS-EXT-STS:vm_state']
+        self.power_state = self.os_rest.response_json['OS-EXT-STS:power_state']
+        logging.info("     %s: %s, power %s",
+                     self.name, self.vm_state,
+                     self.POWERSTATES[self.power_state])
```
edsantiago: It makes me uncomfortable that the POWERSTATES are hardcoded here, with no coordination with the server. If next week they decide to add state 8, this well-intentioned debug statement will cause a difficult-to-diagnose crash.

kommandir/bin/adept_openstack.py
```diff
@@ -0,0 +1,720 @@
...
+            raise
+
+    def service_request(self, service, uri, unwrap=None, method='get', post_json=None):
+        """
+        Make a REST API call to uri, return optionally unwrapped json instance.
+
+        :param service: Name of service to request uri from
+        :param uri: service URI for get, post, delete operation
+        :param unwrap: Optional, unwrap object name from response
+        :param methd: Optional, http method to use, 'git', 'post', 'delete'.
```
edsantiago: get, not git

kommandir/bin/adept_openstack.py
```diff
@@ -0,0 +1,720 @@
...
+        logging.info("Creation successful, attempting to assign floating ip to VM %s", name)
+        floating_ip = TimeoutAssignFloatingIP(server_id)()
+        logging.info("Floating IP assignment successful, ip %s", floating_ip)
+    except:
+        # Fire and forget
+        os_rest = OpenstackREST()
+        os_rest.server_delete(name)
+        raise
+    sys.stdout.write(OUTPUT_FORMAT.format(name=name, floating_ip=floating_ip))
+    return 0
```
edsantiago: It takes some effort to figure out that `0` here is a UNIX exit code which will eventually be passed to sys.exit(). Since none of `create`, `discover`, or `delete` ever return anything other than 0, why bother? Why not just fall through without an explicit return value? That will be less confusing to the reader.

kommandir/bin/adept_openstack.py
```diff
@@ -0,0 +1,720 @@
...
+    except IndexError:
+        raise ValueError("Must pass server name as first parameter")
+    if operation in ['create', 'discover']:
+        if len(argv) < 1:
+            raise ValueError("Must pass server name, and one or more"
+                             " paths to public ssh key files as parameters")
+        parsed_args['pub_key_files'] = set(argv)
+    elif operation == 'help':
+        logging.info("FIXME: Some useful --help message")
+    elif operation not in ('discover', 'destroy'):
```
edsantiago: Why not just `elif operation != 'destroy'`? The 'discover' case is handled above.

edsantiago: Also, doesn't 'destroy' need an argument? So shouldn't it also be in line 575?

cevich:
> OK, I'm done for the day.

Way more than enough, thanks!

> This is much more complicated than I ever expected.

You and me both.

> leftover cruft from previous iterations. I would like for it to be cleaned up before checkin.

Absolutely, I appreciate your sharing of your finely-tuned, expertly-calibrated cruft-detector (mine is totally fubar'd after the unitest grind).

kommandir/bin/adept_openstack.py
cevich: got it 😀

kommandir/bin/adept_openstack.py
```diff
@@ -0,0 +1,720 @@
...
+        start = time.time()
+        if self.time_out_at is None:
+            self.time_out_at = start + self.timeout
+        while True:
+            if time.time() >= self.time_out_at:
+                raise self.timeout_exception(str(self))
+            result = self.action(*args, **dargs)
+            if self.done():
+                break
+        return result
```
cevich: Original reason was to guarantee at least one call to ``action()``.  However, perhaps that's not important anymore, looking to see if your better/simpler way could be realized...

cevich: ...Seems like I could just nuke ``action()`` and move all that code into ``__init__()``.  There are a few subclasses that would need minor re-jiggering, but it could work.  Giving it a try...

edsantiago: Wait, what?

edsantiago: I'm confused. How does my suggestion change anything about `action()` being called?

edsantiago: It's just a two-line change... hold on please don't make huge changes to init or anything, this was a super-duper-trivial change

cevich: In use, ``action()`` isn't needed, that stuff can all just go into ``__init__()``.  i.e. in actual use, there isn't a case of needing to performing some action, more than once.  Like this:

```python
class TimeoutAction(object):
    """
    ABC callable, raises an exception on timeout, or returns non-None value of done()
    """

    timeout = 300.0  # (seconds)
    time_out_at = None  # absolute
    timeout_exception = RuntimeError

    def __init__(self, *args, **dargs):
        """Abstract initialization"""
        raise NotImplementedError

    def __str__(self, *args, **dargs):
        return ("%s(*%s, **%s) after %0.2f"
                % (self.__class__.__name__, args, dargs, self.timeout))

    def __call__(self, *args, **dargs):
        """
        Repeatedly call ``am_done(*args, **dargs)`` until timeout or non-None return
        """
        result = None
        start = time.time()
        if self.time_out_at is None:
            self.time_out_at = start + self.timeout
        while result is None:
            if time.time() >= self.time_out_at:
                raise self.timeout_exception(str(self))
            result = self.done(*args, **dargs)
        return result

    def am_done(self, *args, **dargs):
        """Abstract method, must return non-None to stop iterating"""
        raise NotImplementedError
```

cevich: New (simpler) usage example:

```python
class TimeoutDeleted(TimeoutAction):
    """Helper for to make sure a server is deleted"""

    timeout = 60

    def __init__(self, name):
        super(TimeoutDeleted, self).__init__(name)
        self.os_rest = OpenstackREST()
        self.os_rest.server_delete(name)

    def am_done(self, name):
        """Return Non-None if server does not appear in server list"""
        time.sleep(0.1)  # don't busy-wait
        if self.name not in self.os_rest.server_list():
            return self
        else:
            return None
```

kommandir/bin/adept_openstack.py
```diff
@@ -0,0 +1,720 @@
...
+
+    def done(self, *args, **dargs):
+        """Abstract method, must return False to continue iterating"""
+        raise NotImplementedError
+
+    def action(self, *args, **dargs):
+        """Abstract method, called until condition() == False or timeout reached"""
```
cevich: fixed.

kommandir/bin/adept_openstack.py
cevich: They're hard-coded in the API docs.  Though you're right.  I can fix with something like:

``power_state = self.POWERSTATES.get(server_details['OS-EXT-STS:power_state'], 'UNKNOWN')``

Then the logging message is still human-readable.

kommandir/bin/adept_openstack.py
cevich: good point.  deleting them...

kommandir/bin/adept_openstack.py
cevich: The entire ``parse_args()`` contents is being replaced with use of ``argparse`` in #121

cevich:
@edsantiago see if you like that better, esp. ``TimeoutAction`` and it's sub-classes.  I think it's much simpler.  I haven't touched the unittests yet, but can re-jigger them if you think this way is better / clearer.

kommandir/bin/adept_openstack.py
afomm: Remove "for" from comment "Helper for to make sure a server is deleted"

kommandir/bin/adept_openstack.py
```diff
@@ -0,0 +1,702 @@
...
+        raise NotImplementedError
+
+
+class TimeoutDeleted(TimeoutAction):
+    """Helper for to make sure a server is deleted"""
+
+    timeout = 60
+
+    def __init__(self, name):
+        super(TimeoutDeleted, self).__init__(name)
```
edsantiago: I'm really really confused. How does this not raise `NotImplementedError`?

cevich:
> I'm really really confused. How does this not raise NotImplementedError?

Sorry, should have mentioned:  "All fixed, totally un-unittested".  Just did a ``--force`` push that addresses everything, fixes up unitests, and everything should be golden.  Thanks to your suggestion, the ``TimeoutAction`` base-class changes resulted in about 4-fewer REST api calls per operation (i.e. big speed boost).

kommandir/bin/adept_openstack.py
cevich: Thanks afom, got it.

cevich:
```
$ time kommandir/bin/openstack_discover_create.py foobar SOMEPUBKEY
INFO:root:Setting up python virtual environment under /tmp/SOMEDIR/.virtualenv
INFO:root:Upgrading pip
INFO:root:Installing dependencies
INFO:root:Reactivating python virtual environment
INFO:root:Attempting to find VM foobar.
WARNING:root:Failed to find existing VM: foobar.
INFO:root:Attempting to create new VM foobar.
INFO:root:Deleting any partially created VM foobar
INFO:root:Submitting creation request
INFO:root:     foobar: building, power NOSTATE
INFO:root:     foobar: building, power NOSTATE
INFO:root:     foobar: building, power NOSTATE
INFO:root:     foobar: building, power NOSTATE
INFO:root:     foobar: active, power RUNNING
INFO:root:Creation successful, attempting to assign floating ip to VM foobar
INFO:root:Found disused ip 1.2.3.4
INFO:root:Attempting to assign floating IP 1.2.3.4 to server id NUMBER
INFO:root:Floating IP assignment successful, ip 1.2.3.4
---
inventory_hostname: foobar
ansible_host: 1.2.3.4
ansible_ssh_host: {{ ansible_host }}
INFO:root:Recording all response JSONs into: /tmp/tmp.bY7pI1a76S.adept.workspace/.virtualenv/openstack_discover_create_api_responses.json

real	0m31.470s
user	0m0.898s
sys	0m0.086s
```

cevich:
@afomm @edsantiago Just pushed more spelling / docstring fixes.  Good enough to merge?  (more fixes can go into #121)

edsantiago:
OK by me, but note that I have not actually reviewed anything. It's moving way too quickly and changing way too much. Just merge it then fix whatever breaks as needed.

edsantiago:
One more clarification: my "suggestion" was a two-line change, not a hundred-line rewrite.

cevich:
> OK by me, but note that I have not actually reviewed anything. It's moving way too quickly and changing way too much. Just merge it then fix whatever breaks as needed.

Yep, quick-look is totally fine.  Really it's an optional-script anyway, anybody can replace it with anything they prefer at any time.  

> One more clarification: my "suggestion" was a two-line change, not a hundred-line rewrite.

Lol, yes, but you were right and the result is cleaner and faster, so I'm not complaining 😁

cevich:
rebase...squash...--force push.

PR #121, opened 2017-03-08
Simplify peon provisioning

cevich:
This set of changes is about re-using the kommandir openstack provisioning script (and ansible-technique) to also provision the peon VMs.


cevich:
[(5) (CN) ADEPT's test VM creation rewrite](https://trello.com/c/MMH7mHUR/278-5-cn-adept-s-test-vm-creation-rewrite)

cevich:
**DNR**: Still incorporating improvements over https://github.com/cevich/autotest-docker-enabled-product-testing/pull/115

cevich:
@edsantiago completed conflict resolution / rebasing main code and unittests.  Added a new feature and unittest, for exclusive-creation (i.e. fail if a VM w/ same name already exists).  

Architecture Reminder / Context for this PR (in bold):
* Exekutir runs ``adept.py exekutir.xn``
* Role ``exekutir/roles/kommandir_discovered`` calls ``kommandir/bin/openstack_discover_create.py kommandir_name ssh_pub_key``
* Exekutir ssh's into kommandir VM as user $UUID
* On kommandir VM, in ``/home/$UUID/``, run ``adept.py job.xn``.
* For each peon (test VM) in ansible inventory:
    * Role ``kommandir/roles/to-be-written`` ****calls ``kommandir/bin/openstack_exclusive_create.py peon_name --other --options``****
    * Migrated (old adept) roles do remaining peon setup/configuration work
    * docker autotest is copied, setup, and run
    * Role ``kommandir/roles/also-to-be-written`` calls ``kommandir/bin/openstack_destroy.py`` for all the peons.

Would you mind taking a *quick* (i.e. fast and incomplete) peek while the diff is managable?

Your input is highly valued and helpful!

cevich:
Quick link to main code:  https://github.com/cevich/autotest-docker-enabled-product-testing/pull/121/files#diff-d580606668133bebfaa51db413357274

edsantiago:
I'm sorry, that's just too much to review, especially when I don't know if it's beingcompletely rewritten as I look at it. I think it's just best to check it in and, if anything breaks, deal with it as needed.

Just one comment: should "homdir" be "homedir"? Does it matter? (in the userdata yaml)

cevich:
Nope, just a quick peek...and you've done it...thanks 😀 

> Just one comment: should "homdir" be "homedir"? Does it matter? (in the userdata yaml)

Ohh, that is a good catch.  I'm not sure, that's cloud-init / cloud-config spec.  I'll look it up again and make sure.  Thanks!

cevich:
> beingcompletely rewritten as I look at it

Nope, I've been off doing other things.

Yep, it is suppose to be spelled correctly, ref: http://cloudinit.readthedocs.io/en/latest/topics/modules.html#users-and-groups  Awesome.  Fixing...

cevich:
@edsantiago Mind taking a _quick_ peek at [adept_openstack.py](https://github.com/cevich/autotest-docker-enabled-product-testing/blob/d30c0eef4a29dac48b758791ddac4b8dd9a3a0df/kommandir/bin/adept_openstack.py) for typos / other simple stuff?  **Not** a full review, just quick sanity check.

Oh, and the (unittest code)[https://github.com/cevich/autotest-docker-enabled-product-testing/blob/d30c0eef4a29dac48b758791ddac4b8dd9a3a0df/test_openstack.py] also.

Please and thank you 😀 

Soon as I can merge this PR, my plan is to move on to porting/migrating the ``fixme_migrate_roles`` over to the simplified way.  So this PR is basically the second-to-last stop for the choo-choo (at least until the final testing and beating).

I've run the script by-hand on my rhel/centos bastard-distro, with various command line options, and it seems to behave.  Still haven't tried running it on Fedora, but that may just be out of scope to support (moves too quickly).

```
$ export WORKSPACE=/tmp/tmp.4pFY7TJRRH/
$ kommandir/bin/openstack_discover_create.py --flavor m1.tiny --router somenetwork --private --userdata /dev/null --timeout=10 --size=10 foobar ~/.ssh/id_rsa.pub 
INFO:root:Setting up python virtual environment under /tmp/tmp.4pFY7TJRRH/.virtualenv
INFO:root:Upgrading pip (inside virtual environment)
INFO:root:Installing dependencies (inside virtual environment)
INFO:root:Reactivating python virtual environment
INFO:root:Processed arguments: {'userdata_filepath': '/dev/null', 'name': 'foobar', 'router_name': 'somenetwork', 'image': 'CentOS-Cloud-7', 'private': True, 'pub_key_files': ['/home/me/.ssh/id_rsa.pub'], 'flavor': 'm1.tiny', 'size': 10}
INFO:root:Attempting to find VM foobar.
WARNING:root:Failed to find existing VM: foobar.
INFO:root:Processed arguments: {'userdata_filepath': '/dev/null', 'name': 'foobar', 'router_name': 'somenetwork', 'image': 'CentOS-Cloud-7', 'private': True, 'pub_key_files': ['/home/me/.ssh/id_rsa.pub'], 'flavor': 'm1.tiny', 'size': 10}
INFO:root:Attempting to create new VM foobar.
INFO:root:Deleting any partially created VM foobar (if any exist)
INFO:root:Confirmed VM foobar does not exist
INFO:root:Submitting creation request for foobar
INFO:root:     foobar: building, power NOSTATE
INFO:root:     foobar: building, power NOSTATE
INFO:root:     foobar: building, power NOSTATE
INFO:root:     foobar: building, power NOSTATE
INFO:root:     foobar: active, power RUNNING
INFO:root:Creation successful, VM id c6ebd098-608c-489e-83bc-c31d511cd2d4
INFO:root:Attempting to create and attach 10GB size volume
INFO:root:Creating 10GB volume named foobar
INFO:root:     foobar: status available attachments: []
INFO:root:     foobar: status attaching attachments: []
INFO:root:     foobar: status attaching attachments: []
INFO:root:     foobar: status in-use attachments: [{u'server_id': u'c6ebd098-608c-489e-83bc-c31d511cd2d4', u'attachment_id': u'9b483f38-f474-47f5-8927-1bc16b10fd07', u'host_name': None, u'volume_id': u'ad885885-29dd-4fbd-b660-5c8c9688d9c3', u'device': u'/dev/vdb', u'id': u'ad885885-29dd-4fbd-b660-5c8c9688d9c3'}]
---
inventory_hostname: foobar
ansible_host: 1.2.3.4
ansible_ssh_host: {{ ansible_host }}
INFO:root:Recording all response JSONs into: /tmp/tmp.4pFY7TJRRH/.virtualenv/openstack_discover_create_api_responses.json
```

edsantiago:
Compared (manual diff) against earlier version. I don't know how to leave review feedback.

In `server_delete()` I don't understand the change from 'server' to 'server(s)'. AFAICT `name` is never interpreted as a list; and I don't see how there can be multiple servers with the same name.

In `volume`, I'm uncomfortable with the null call to `self.volume_list()` and then using side effects. 

In `TimeoutCreate.__init__()`, could you reverse the userdata_filepath conditional? I find that style unreadable. I find this more readable:

    if userdata_filepath:
        ...then read it
    else:
        ...default to something hardcoded

In `TimeoutAttachVolume`, eek, do you really want to have volumes with the same name as the server? No `_vdb` or `_vol` or anything? I would find it confusing to look through the GUI or run openstack commands and see 'myserver' both as a system and a volume.

In the new `parser.add_argument()` there's a hardcoded 300 that should be DEFAULT_TIMEOUT.

Just pass 1; I skipped the new DeleteVolume, will look at that later.

cevich:
@edsantiago thanks for the quick look.  Thoroughness-level is perfectly adequate, I'll resume work on this once I'm finished looking at that silly ``docker_cli.run_volumes_selinux`` test...

cevich:
> and I don't see how there can be multiple servers with the same name.

Unf. OpenStack allows this (clashing names), the UUID is the only unique thing.  However ADEPT strives to rely on the name being unique, because it's so much easier to work with (e.g. in playbooks/inventory) as it becomes the system's hostname.  This also allows naming the hosts based on the job/run that created them, for "WTF did this VM come from" tractability.

So name-clashes are possible and there's no API to guarantee created VMs have both unique and human-readable names.  For example, if multiple humans share an OpenStack tenant with automation, it's possible to accidentally create duplicate named VMs.  Especially if the humans are working with a tool from a shared codebase that has a default name encoded (e.g. ``adept_kommandir``) 😀 

The exact same problem exists with volumes, but it's worse because the script strives to have volume names match system names.  This means you can have a leaked volume with a name that matches a yet-to-be-created system name.  Though I think there's a fix for this (see below)

Possibly deleting VMs some other script created makes me a bit nervous as well.  Perhaps the way to address your concern (about VM names), is to revert ``server_delete()`` to be singular-only.  Then in the script, upon discovery of any name clashes, nuke any thus-far created servers/volumes by ID (from the current execution), and bail with an error.

Does that make more sense and/or sound reasonable?

> In volume, I'm uncomfortable with the null call to self.volume_list() and then using side effects.

Gotcha, no prob, easy to fix.

>  reverse the userdata_filepath conditional?

Yes, good catch.

> there's a hardcoded 300 that should be DEFAULT_TIMEOUT

Ohhh, good catch, thanks!

> do you really want to have volumes with the same name as the server?

I found it easier to read the listing this way, esp. if a volume has been leaked (no matching server).  But to each their own.  I could just use the volume's ID as it's name, and that would eliminate the clashing problem for them.  There is a description field, I could stick the intended attachment server name into that (for human reference).  Okay, I'll look into it.

cevich:
@edsantiago Okay, I think I addressed everything you brought up.  (New fixup! commit).  I haven't touched up the unittests, wanted to see if you like the look/feel of this better now:  https://github.com/cevich/autotest-docker-enabled-product-testing/blob/699052a6ffb2a535b7306d30676b7c733d874944/kommandir/bin/adept_openstack.py

* Changed volumes to use the uuid of their intended target VM.
* Add intended target VM and it's ID to new volume's description
* Tweaked ``create()`` and ``delete()`` so they work internally with ID's and bail out if a name clash is detected.
* Tweaked some ``OpenstackREST()`` methods to support more uuid-centered operations.

edsantiago:
Sorry for the late response. That was a lot to review.

I'm really uncomfortable with `TimeoutDeleted()`, its name and its undocumented (and unexpected) return value. But I'm reluctant to gripe much about it because I know it might lead to more huge changes and these are really hard for me to review. Can you just check it in, see if it works, and just deal with fixes as needed?

cevich:
> undocumented (and unexpected) return value.

Undocumented I can fix pretty easy, and that should make the return value more clear.

> deal with fixes as needed?

I'm fine with that, but it implies both of these need to be supportive of future updates/changes/fixes:
*  The level of indirection / modularity
*  The interfaces between major components (parameters)
*  The top-level names

Assuming the first two are adequate, now is the best time to do something about any names you'd like changed.

How about, ``TimeoutDeleted()`` -> ``TimeoutGetVolumesDelVM()`` ?

Alternatively, I could try to dis-entangle the two (makes the exception-handling more hairy).

cevich:
@edsantiago Okay, how about this (I think it's cleaner):  https://github.com/cevich/autotest-docker-enabled-product-testing/pull/121/commits/545f49a3c394dc8865a3fbe9e47355a6863fe19f
*  ``TimeoutDeleted()`` -> ``TimeoutDelete()``
*  Fixed funky return value
*  Re-use ``destroy()`` in case of exception during ``create()``.

Still need to update unitests, but that normally goes pretty quickly.

edsantiago:
> it implies both of these need to be supportive of future updates/changes/fixes

Is that out of respect for potential users of this code? If so, _are_ there any? Even assuming there are, could I ask you to consider whether they might accept a little instability? This is important to me. I find the concept of "get it right the first time" paralyzing. It is obviously impossible in practice. 

If it's another reason, please let me know what I'm missing. Thanks.

cevich:
> get it right the first time

No other users and I don't really care...I don't mean _all_ the code, just mean the birds-eye-view, major stuff.  

Locked-in function/method signatures/returns, and a good level of modularity helps with writing/maintaining unittests and compatibility with external callers (i.e. playbooks).   It allows extending and fixing stuff w/o having to traipse all over, redesigning, and **constantly** reaching for the duct-tape.

I don't want it perfect, I just want (low-viscosity)[https://en.wikipedia.org/wiki/Viscosity_(programming)] (some is inevitable, I just don't want excess).  IMHO, the viscosity is "okay" as-is now.  The only thing I question ATM, is the choice of return-values for the ``Timeout*()`` classes (as you rightly pointed out before).  So, do you think low-ish enough viscosity, or spot anything last minute that could use a quick-fix or rename?

If 'good-enough', then I'll get back to fixing the unitetsts, then merge, and move along.

edsantiago:
The query-volumes, then delete, then delete volumes approach is much cleaner. Thank you.

I don't understand the sentinel thing at all, and am giving up. The rest of the changes look fine.

cevich:
> I don't understand the sentinel thing at all, and am giving up.

It's there purely due to my ignorance of a better way.  That's low-level stuff though, we can nuke it w/o much affect.  Okay, working on unittests then, thanks for your help...

cevich:
Unitests all passing, even added a few new ones and fixed a few bugs they found.

PR #122, opened 2017-03-17
Experiment: Don't install docker-selinux, it's obsoleted by container-selinux

cevich:
In some flavors of fedora/rhel/centos, package dependencies and/or
obsolete-fixing is broken such that; if installed, every 'docker run'
command will trigger avc denials + go panics.


cevich:
No difference.

PR #123, opened 2017-03-20
Support upgrade testing

cevich:


cevich:
[(2) [CN] RHEL upgrade testing in CI](https://trello.com/c/id2OeNeL/440-2-cn-rhel-upgrade-testing-in-ci)

cevich:
Note: This won't actually DO upgrade testing by default, ``variables.yml`` overrides will need to be added by the caller to define those peons.

PR #125, opened 2017-03-21
Two small tweaks

cevich:


cevich:
[(2) [CN] RHEL upgrade testing in CI](https://trello.com/c/id2OeNeL/440-2-cn-rhel-upgrade-testing-in-ci)

cevich:
CI Fails because it's trying to spin up CentOS peons but those peons aren't defined anymore.

PR #126, opened 2017-03-21
Fix version number vs string

cevich:
This value is used for looking up the image when creating a peon.
However, when used as mapping keys, the object type is important.
This supports simple typos like treating a two-component version
as if it was a floating point number.  Since the image map uses
string version number keys, the floating-point numbers from the peon
definition wouldn't match, resulting in an error.  This commit
both fixes the source of the problem, and adds a conversion
of all peon version values, into strings.


cevich:
Note: Testing this internally since it depends on some other supporting changes.

PR #130, opened 2017-04-11
Update Timeouts and Fedora Test Image

cevich:


cevich:
[(1) [CN] Fix atomic timeouts and fedora test image](https://trello.com/c/V7bCkiFV/452-1-cn-fix-atomic-timeouts-and-fedora-test-image)

PR #132, opened 2017-04-13
typo fix: registry.access.redhat.com

edsantiago:
Should fix this ADEPT failure in docker create test:

   Get https://registery.access.redhat.com/v1/_ping: dial tcp: lookup registery.access.redhat.com on 172.16.12.2:53: no such host: expected 0; got 1


cevich:
Woops, good catch!  Thanks!

PR #134, opened 2017-04-25
Fix selinux permissive

cevich:


roles/atomic_upgrade/tasks/main.yml
```diff
@@ -35,9 +35,8 @@
 - block:
 
     - name: Workaround BZ1290281
-      selinux:
-        policy: targeted
-        state: permissive
+      # Only temporary, don't touch /etc/sysconfig/selinux
```
edsantiago: As long as you're touching these lines, how about (1) an explanation of the bug, something like "atomic host upgrade barfs with fsetxattr error", and (2) double-checking and maybe fixing the bz number, since I don't think that's the right one?

roles/atomic_upgrade/tasks/main.yml
```diff
@@ -35,9 +35,8 @@
...
+      command: "setenforce 0"
```
edsantiago: why the quotes?

roles/atomic_upgrade/tasks/main.yml
```diff
@@ -58,9 +57,8 @@
   always:
 
     - name: Undo Workaround BZ1290281
-      selinux:
-        policy: targeted
-        state: enforcing
+      # Just in case any subsequent roles/tasks depend/check on this
+      command: "setenforce 1"
```
edsantiago: ibid & ibid

roles/atomic_upgrade/tasks/main.yml
cevich: it's the right bug, well one of them at least.  I don't understand how os-tree / atomic works well enough to explain/comment about the bug.  All I know is that's the one in play here (for older atomic hosts).  If you understand better or can suggest a comment, I'll happily add one.

roles/atomic_upgrade/tasks/main.yml
cevich: Just a habit.

roles/atomic_upgrade/tasks/main.yml
```diff
@@ -34,10 +34,9 @@
 
 - block:
 
-    - name: Workaround BZ1290281
-      selinux:
-        policy: targeted
-        state: permissive
+    - name: Workaround old atomic command bug: failure b/c /tmp is a symlink
```
cevich: @edsantiago marginally better?

edsantiago: Not really... as a monkey who's simply on the user side of things, I don't know or care about /tmp being a symlink. All I know is that when I run atomic upgrade, I see a barf message with something about fsetxattr, and the solution is `setenforce 0`.  At least, I think that's why you're doing this, right? Anyway, the point is, if a future maintainer decides to delete this code, and the fsetxattr barf reappears, having "fsetxattr" in the description would be a really helpful way to realize that it's not yet time to delete it.

edsantiago: TL;DR I like describing symptoms, not nitty-gritty details

cevich: TBH, I dunno if that's even the same error we see on RHEL.  Having a misleading comment is more unhelpful than having no comment.  Pretty sure it does say "/tmp has too many levels of symlinks".  You think it should be like:

``Workaround fsetxattr bug in older atomic command versions``

edsantiago: I guess I'll stand down. I don't know what problem this is intended to fix; it might be different from the one I'm thinking of. It's not worth spending more time on. Sorry for the confusion.

cevich: No worries, there probably isn't a good way to word it when neither of us really understands the underlying problem 😁

cevich:
Eww, CI found a problem here:

```

13:45:19 ERROR! Syntax Error while loading YAML.

13:45:19 

13:45:19 

13:45:19 The error appears to have been in '/var/lib/adept/roles/atomic_upgrade/tasks/main.yml': line 37, column 46, but may

13:45:19 be elsewhere in the file depending on the exact syntax problem.

13:45:19 

13:45:19 The offending line appears to be:
13:45:19 
13:45:19 
13:45:19     - name: Workaround old atomic command bug: failure b/c /tmp is a symlink
13:45:19                                              ^ here
13:45:19 
```

cevich:
[(2) [CI/CN] Fix selinux: permissive in ADEPT and Docker Autotest](https://trello.com/c/H7hlA8at/464-2-ci-cn-fix-selinux-permissive-in-adept-and-docker-autotest)

PR #135, opened 2017-04-26
Update control file template for autotested role

cevich:
After autotest/autotest-docker #710 it's not necessary to re-specify all
non-default control.ini options.  Only those which should override them
need to be written.


cevich:
Note: This depends on https://github.com/autotest/autotest-docker/pull/710

PR #140, opened 2017-05-23
Fix Fedora/pull subtest

cevich:
It was incorrectly trying to pull the built test-image instead of the
base image.  However, while editing the template anyway, just split the
entire thing by the ``is_enterprise`` value.  This makes it clearer
which sections apply to which platforms.


cevich:
jenkins, test this please.

PR #141, opened 2017-05-31
Don't kill the tool needed to do the work

cevich:
At the end of cleanup, some directories in the kommandir's workspace are
removed after sync'ing them back to the starting host.  This partly
prevents rapid consumption of workspace cache on the kommandir VM.
However, it was removing the ``adept`` directory, which is required by
external callers to perform additional cleanup steps.  Don't remove this
directory.


cevich:
jenkins, test this please.

PR #142, opened 2017-06-09
Fix liverestore peon configuration

cevich:
Based on RHEL ``7.4`` testing, a missing option in the docker/latest unit
files was found.  With this commit, the only special/additional setup
steps for enabling liverestore are:

* Starting with daemons stopped, ``container-storage-setup``
  hasn't run yet.
* Rely on runc package to have tweaked kernel params (all platforms)
* Copy ``/usr/lib/systemd/system/docker*.service`` into
  ``/etc/systemd/system/``
* In the copies, replace ``MountFlags=slave`` with ``KillMode=process``
* Start/enable docker or docker-latest services (implied
  ``container-storage-setup``)

### DIFF
--- a/roles/liverestore_enabled/tasks/main.yml
+++ b/roles/liverestore_enabled/tasks/main.yml
@@ -19,11 +19,11 @@
   with_fileglob:
     - "/usr/lib/systemd/system/docker*.service"
 
-- name: The MountFlags=slave option is not present in customized unit files
+- name: The MountFlags=slave option is replaced with KillMode=process in unit files
   replace:
     dest: "{{ item }}"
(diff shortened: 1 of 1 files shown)


edsantiago:
It might be a lot simpler to just create `/etc/systemd/system/docker.service.d/liverestore.conf`:

    [Service]
    KillMode=process

cevich:
@edsantiago see! That's exactly why I need your help here.  I had no idea that was even possible!

cevich:
I wish you could do a conditional-loop with ``block:`` 😖 

I s'pose the alternative here is to deploy a jinja2 template.

PR #150, opened 2017-07-10
Fix not using global lock on kommandir

cevich:
On an active Kommandir, it's possible for jobs to contend with eachother
over allocating floating IP addresses.  Partially address this by
using a system-wide lock directory.  This was previously setup in a
prior commit, but not realized via kommandir/inventory/group_vars/openstack.yml

This does not address the issue where multiple kommandirs all contend
for public IPs, however solving that is non-trivial and out-of-scope for
this small fix.

Also include a fix to adept_openstack.py that results in CWD being
used as the lock dir because of default '' (empty string) parameter value.


cevich:
jenkins, test this please.

PR #152, opened 2017-07-11
Fix & Simplify workspace_cleanup

cevich:
Previously some rather complicated list-manipulations were needed to
essentially delete a set of files/dirs that are really only known
to the caller.  Simplify this role to do nothing by default, unless
the caller specifies a cleanup_globs list.  Only in that case,
remove matched files from the workspace that aren't also in the
cleanup_filter list (e.g. the workspace itself).

Also update the exekutir_workspace_setup such that the formed
``kommandir_vars.yml`` simply includes the same
cleanup_globs/cleanup_filter as used on the exekutir.


cevich:
jenkins, test this please.

cevich:
jenkins, test this please.

PR #157, opened 2017-07-13
New docker vers. need rpm for deferred removal

cevich:


cevich:
This will get rebased from legacy adept into simplify

PR #167, opened 2017-08-02
Fix use of namespace for fedora_test_image

cevich:
If it's there, then docker feels compelled to try and contact the
localhost registry (which doesn't exist). Instead, if the
default test image was deleted, we want tests to fail as early as
possible.

Also updated documentation. This change needs to be paired with
a similar change being made in Docker Autotest


cevich:
https://github.com/autotest/autotest-docker/pull/749

PR #169, opened 2017-08-15
Support passthrough of --args to Docker Autotest

cevich:
* Secure passing of external variable values, by filtering them after
parsing.  This prevents a multi-line incoming string value from
corrupting or leaking/overwriting other values.  Previously this was
possible by blindly using 'lineinfile' to set YAML values in
kommandir_vars.yml

* Define new variable ``pull_request_description`` and add
documentation.  Include this value from ``exekutir_vars.yml``
when producing ``kommandir_vars.yml``.

* Define a new variable, ``docker_autotest_mmargs`` for the Autotested
role.  Document this in the role's defaults ``main.yml`` file.
When set, it's contents will be passed to the ``--args`` option
of autotest-local via an environment variable.  The indirect reference
blocks attempts to escape the confines of quoting.

* Add safe-parsing of a magic string from pull_request_description
by first removing white space and unacceptable characters.  When the
filtered result contains ``[--args=<foo>]``, the '<foo>'
value will be extracted and become ``docker_autotest_mmargs``.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,6 @@ script:
     - cd docs && SPHINXOPTS="-W" make html
     - cd ..
     - unit2 -vfc
+    - ./test_exekutir_xn.sh
     # Check for some typos / git commit problems (except in this file)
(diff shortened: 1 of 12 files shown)


cevich:
jenkins, test this please.

cevich:
[(1) [CN] Docker Autotest CI Optimization](https://trello.com/c/xRqq6XzW/280-1-cn-docker-autotest-ci-optimization)

PR #175, opened 2017-09-13
autotested: Copy known_failures.txt

cevich:
A disposable PR to help test https://github.com/autotest/autotest-docker/pull/761 because I'm lazy and probably screwed up the ``known_failures.txt`` file somehow...

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -93,7 +93,7 @@ Latest Documentation
 ======================
 
 For the latest, most up to date documentation please visit
-http://ADEPT.readthedocs.io/en/latest
+http://A-D-E-P-T.readthedocs.io/en/latest
 
 The latest `Docker Autotest`_ documentation is located at:
 http://docker-autotest.readthedocs.io
--- a/kommandir/roles/autotested/files/known_failures.txt
+++ b/kommandir/roles/autotested/files/known_failures.txt
@@ -1,0 +1,1 @@
+# <NVRA> <subtest path> <human-friendly description>
--- a/kommandir/roles/autotested/tasks/setup.yml
+++ b/kommandir/roles/autotested/tasks/setup.yml
@@ -72,6 +72,11 @@
 
 - include: configure_control.yml
 
+- name: Copy 'known_failures.txt' file to config_custom directory
+  copy:
+    dest: "{{ autotest_docker_path }}/config_custom/"
+    src: "known_failures.txt"
+
 - name: Results directory is empty
   file:
   args:


jobs/adept_pr/known_vailures.txt
```diff
@@ -0,0 +1,5 @@
+# <NVRA>    <subtest name>   <human-friendly description>
+docker-1.12.6-55.gitc4618fb.el7.x86_64     docker_cli/run_volumes   was     rhelatomic7-4-docker.subtests.docker_cli.run_volumes
+docker-1.13.1-19.git27e468e.fc26.x86_64            docker_cli/run_volumes   was fedoratomic26-docker-latest.subtests.docker_cli.run_volumes
+docker-*      docker_cli/systemd_in_container     everywhere
+docker-latest-42-*      docker_cli/version                  nowhere
```
cevich: N/B: I just put this file here for now, it's permanent home can be elsewhere.

jobs/adept_pr/kommandir_vars.yml
```diff
@@ -1,7 +1,9 @@
 ---
 
-docker_autotest_timeout: 30
+docker_autotest_timeout: 60
 
-job_subthings:
-    - docker_cli/dockerhelp
-    - docker_cli/version
```
cevich: Killing that (above) should cause all tests to run.

cevich:
OMG! I can't even spell a filename correctly...

cevich:
(travis failure is just due to the ``fixup!``)

cevich:
once again for the third time

cevich:
okay, known_failures is being copied, but something else is breaking, perhaps the file is formatted wrong?

edsantiago:
What do you mean by "something else is breaking"? What are you expecting?

At a first guess from looking at the diffs and the ADEPT results: I see failures in `run_volumes/oci_umount` but the known_failures lines I see are just `run_volumes`. Is it your intention to hide the oci_umount errors? If so, try `docker_cli/run_volumes/oci_umount`

cevich:
I was expecting:

* Any ``run_volumes`` sub-subtest that failed on a ``{rhel,fedora}atomic`` peon should get filtered, but should fail properly on plain rhel/fedora.  
* The ``docker_cli/systemd_in_container`` errors should get filtered on ALL peons.  
* The docker_cli/version test shouldn't be touched anywhere

jobs/adept_pr/roles/autotested/files/known_failures.txt
```diff
@@ -0,0 +1,8 @@
+# <NVRA>                                   <subtest name>           <human-friendly description>
+docker-1.12.6-55.gitc4618fb.el7.x86_64     docker_cli/run_volumes   should get filtered on rhelatomic7-4-docker peons
```
edsantiago: You can't filter all sub-subtests by giving a subtest. You have to say `docker_cli/run_volumes/oci_umount`

jobs/adept_pr/roles/autotested/files/known_failures.txt
```diff
@@ -0,0 +1,8 @@
...
+
+docker-1.13.1-19.git27e468e.fc26.x86_64    docker_cli/run_volumes   should get filtered on fedoratomic26-docker-latest peons
+
+docker-*                                   docker_cli/systemd_in_container should be ignored everywhere
```
edsantiago: Wildcard only works for R, not V. E.g. docker-1.12.6-*. Anything like `docker-*` is probably a long-term problem and best handled in ADEPT or docker-autotest itself.

jobs/adept_pr/roles/autotested/files/known_failures.txt
```diff
@@ -0,0 +1,8 @@
...
+
+docker-latest-42-*                         docker_cli/version       should run normally everywhere
```
edsantiago: docker-latest-42??? Is there a versioning change I'm not aware of?

cevich: lol, nope, just negative testing.

jobs/adept_pr/roles/autotested/files/known_failures.txt
cevich: Gotcha.  So if docker and docker-latest ever have the same version (assuming that's possible), we can just list it twice if there's an issue. Kk, np so long as that's documented.

jobs/adept_pr/roles/autotested/files/known_failures.txt
cevich: Hmmm, okay, I can live with that so long as it's documented 😀

cevich:
*sigh* dunno what's I'm doing wrong, seems there's some hidden problem killing-off ``fedora_testimage`` 😖 Tried spotting it in the debug logs, but no love.  Could maybe just make a damn image for it and upload to docker, I dunno.

edsantiago:
I don't see the word 'conflict' in `artifact/results/fedoratomic26-docker-latest/docker_autotest/debug/client.DEBUG`, and the errors I see have to do with pulling, but since you use the term 'killing-off' I'm posting this just in case: there is a problem with my latest docker-autotest PR. I'm trying to track it down right now but having trouble reproducing it. Basically, though, removing an image by ID doesn't work if the same image has multiple tag names. The error is:

    Error response from daemon: conflict: unable to delete 61b442687d68 (must be forced) - image is referenced in multiple repositories

I'm preparing a followup PR but need to figure out how to test it. Expect it tomorrow. ITM I don't think this has anything to do with what you're seeing (again, I don't see the word "conflict") but am posting this just in case I missed something.

cevich:
Hmmm, could be.  Yeah, the multiple name thing, I think I remember that being a major struggle.  In the fedoratomic logs (here) all I saw was one test had the image, the next one didn't.  No indication of where it went and I ran out of patients.  Will look more today.

PR #184, opened 2018-01-03
Add workaround for OpenStack networking bug

cevich:
In the reccomended configuration for OpenStack, some
container-networking opterations can fail due to docker or the kernel
not properly configuring the MTU properly.  This is a long-standing
problem and as of this commit, still being worked on.  As a workaround
for all cases, force the docker and docker-latest MTU size for
containers to 1400 (recommended configuration value).  Add an option
to easily disable this workaround in the future if needed.  Otherwise
leaving this in is basically inconsequential.


cevich:
Note to me: This needs manual testing

PR #193, opened 2018-01-24
Fix oversharing of env. vars

cevich:
Originally, ``adept.py`` did not have the ability to allow selecting or
setting env. vars within a transition file.  Therefor the default
set of variables selected for pass-through was quite wide.

Since adding a ``variable`` action-item to the standard, it's now
possible to select env. vars for pass-through and/or set static values.
The more general set of default variables in ``adept.py`` can therefor
be safely paired back to a more sane minimum.  Including only variables
which are extremely common and safe to allow in by default.


cevich:
note to me: looks like an ansible-bug

PR #194, opened 2018-01-26
Set OpenStack metadata preserve on create

cevich:
This is handy for implimenting 'reaper' based cleanup strategy.
Establish a metadata key ``preserve`` for every created server,
with a default value of ``False``.  If a VM needs to be preserved,
then the openstack command or API can be used to easily flip
this key ``True``.


cevich:
[run ci]

PR #199, opened 2018-03-14
Re-use @edsantiago playbooks/roles for a Basic DaT run

cevich:
@edsantiago PTAQL (quick) when you have a moment.   The commit descriptions should pretty-well summarize what's been done.  The diffs are likely to be a bit overwhelming to look at, so I suggest concentrating your gaze under ``jobs/basic/...`` for now.

My intention here is to get a basic, simple, setup in place for a periodic basic-job, and also have some of the machinery be something that's familiar to you :grinning:   I've tested these changes (internally) and they successfully provision, setup + install DaT on RHEL 7.5-nightly.

TODOs / Uncertainties:

* I had to make a few small changes to your dat-setup playbook/roles, to protect internal-services, and do some minor optimizations (for next bullet).  Though I'm not glued to any of them, everything's negotiable.

* I'd like to get this working with an atomichost peon, and this can be done, but will require some extra fiddling so I left it out for now.  I think getting it to work will just entail grouping some tasks based on inventory-groups (which are already in place).

* I like the idea of the webserver, but I commented it out b/c nginx is only in EPEL which I've stripped out for simplicity-sake.  There is room and a place to add EPEL back into the mix, it's just a SMAP.

* This PR does not make any VM-creation (provisioning) changes.  Keeping that part flexible is one of the central tenants of ADEPT, swapping it up later, will be friction-free as far as job-content is concerned.  I have another PR in the works that will enable reaper-style VM management.


edsantiago:
@cevich I'm sorry, I can't. It would take me days just to make one solid pass on all that; even just `jobs/basic` would take a good part of a day. If you get something working that takes over my cron farm, and you want to take over all the monitoring, I'll be very happy. But I'm really sorry, I just don't have the mental capacity to deal with so much ansible.

cevich:
@edsantiago Wups, that's my miscommunication then: You should find ``jobs/basic/...``is 90% familiar (you wrote it).  All I'm asking for is a peek at the 10% difference.  I'd be surprised if it took you 15-minutes.  Remember, I've already tested it, and it works.

cevich:
@edsantiago ...as suspected...it's a 93-line diff, excluding the items mentioned in my 'protect internal details'  bullet:  https://url.corp.redhat.com/49c3cd2

PR #212, opened 2018-05-02
Add and implement --dry-run option

edsantiago:
The --verbose option was until now being used incorrectly
to mean "show me what you would've done, but don't do it."
This corrects that.

Also, fix some misspellings.


cevich:
...yeah the unittest for ``--verbose`` (``test_openstack.py``, line 597, data-file ``.test_openstack_TestReap.test_verbose.json``) should be failing (or the test is broken).  The good news though is once fixed, the test and data-file can just be renamed for ``dryrun``.

cevich:
Oh and because it's not obvious: Those unittest data files are the mocked JSON responses that would come back from OpenStack REST API, in order (this is the nightmare part of the unittests).

cevich:
(no joke, it's REALLY hard to debug the unit-tests)

cevich:
...idea: Looks like the fix to the unittest may be easy:  The data-file has the bug:  It's checking that no VMs are removed, however all the VMs metadata says ``'preserve': '-1'``.  That just needs to be fixed to be ``1`` instead, along with swapping ``verbose=true`` to ``dry-run=true``, renaming the unittest and data-file.  I think that should fix it and get coverage for the changes.  Make sense?

cevich:
ugg, looks like the ``.test_openstack_TestReap.test_forever_good.json`` symlink needs to be changed to a file (with the ``preserve: -1``).  Gosh, who wrote this crap!

cevich:
@edsantiago I think this should fixup the unittest, or at least get it really close: [patch.diff.txt](https://github.com/cevich/ADEPT/files/1968375/patch.diff.txt)

edsantiago:
@cevich thanks for the unit test fixes - there's no way I could've done those. New changes force-pushed, tests pass.

cevich:
(oh god, that smiley and thumbs-up did NOT align in a good or intended way, on my screen)

PR #215, opened 2018-05-03
Add new verbose logging level filter

cevich:
This actually implements --verbose, by defining two levels
of INFO messages:

    logging.info("This is always displayed")
    logging.info(">Shown only if --verbose")

That is: if message begins with '>', it is filtered out
by default but shown if --verbose (or --debug). The
leading '>' is stripped.

I considered reversing the sense of the test: filtering
out INFO messages by default, and showing only those
starting with '!' or '*' or something. I think that
would be counterintuitive, so I went with show-unless->
even though that makes this diff larger.

### DIFF
--- a/kommandir/bin/adept_openstack.py
+++ b/kommandir/bin/adept_openstack.py
@@ -597,7 +597,7 @@ class TimeoutDelete(TimeoutAction):
         """Return remaining ids when server_id not found, None if still present."""
         server_ids = self.os_rest.server_list(key='id')
         if server_id in server_ids:
-            logging.info("    Deleting %s", server_id)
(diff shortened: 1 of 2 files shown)


cevich:
@edsantiago okay, this is working now and passing all the tests I could throw at it.  PTAL in case I goofed something, though it should be really really close to the work you did.

edsantiago:
LGTM but I can't make sense of the CI failure

cevich:
Ahh, it's just missing an indentation space, got it.

cevich:
okay, that should do it.  I also split out the few small changes I made into separate commits.

PR #217, opened 2018-05-07
Tests for VerboseFilter

edsantiago:
Make sure that the new VerboseFilter works as expected, and
does not inadvertently affect any levels other than INFO

### DIFF
--- a/test_verbosefilter.py
+++ b/test_verbosefilter.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python
+
+"""
+Unit tests for VerboseFilter, a logging helper in adept_openstack.py
+"""
+
+import logging
+import sys
+from mock import Mock
+from unittest2 import TestCase, main
+
+sys.modules['virtualenv'] = Mock()      # module not needed for this test
+
+sys.path.insert(0, 'kommandir/bin')
+
+
+class MyHandler(logging.Handler):
+    """
+    Output handler for logging messages. No output; just push onto list.
+    """
+    def __init__(self, message_list):
+        logging.Handler.__init__(self)
+        self.message_list = message_list
+
+    def emit(self, record):
+        self.message_list.append(record.msg)
+
+
+class TestVerboseFilter(TestCase):
+    """
+    General test class. Defines only one test function, used by
+    test generator to create tests for all combinations of levels
+    and verbose flags.
+    """
+
+    def setUp(self):
+        import adept_openstack  # pylint: disable=E0401
+        self.a_o = adept_openstack
+        self.logger = logging.getLogger()
+        self.message_list = []
+
+    def _test_basic(self, level, verbose, expect):
+        # Grrr. The Logger thing is a global, and there doesn't seem to be
+        # a way to get a new one each time. For addHandler() and addFilter()
+        # to work, we need to delve into internals.
+        self.logger.handlers = []
+        self.logger.filters = []
+
+        # Now we can set up our desired handler and filter
+        self.logger.setLevel(level)
+        self.message_list = []
+        self.logger.addHandler(MyHandler(self.message_list))
+        self.logger.addFilter(self.a_o.VerboseFilter(level, verbose))
+
+        # Always log the same messages...     pylint: disable=C0326
+        logging.debug( "this is DEBUG")
+        logging.info(  "this is INFO")
+        logging.info( ">this is VERBOSE-INFO")
+        logging.warn(  "this is WARN")
+        logging.error( "this is ERROR")
+
+        # ...the difference is what we expect to see.
+        expect_list = []
+        for level_name in ['DEBUG', 'INFO', 'VERBOSE-INFO', 'WARN', 'ERROR']:
+            if level_name[0] in expect:
+                expect_list.append('this is {}'.format(level_name))
+
+        self.assertEqual(self.message_list, expect_list)
+
+def test_generator(test_info):
+    """
+    Generate a test for this combination. Returns test name and code ref.
+    """
+    (level, verbose, expect) = test_info
+
+    name = 'test_' + logging.getLevelName(level).lower()
+    if verbose:
+        name += '_verbose'
+
+    def _run_test(self):
+        self._test_basic(level, verbose, expect.strip()) # pylint: disable=W0212
+
+    return [name, _run_test]
+
+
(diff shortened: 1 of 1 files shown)


cevich:
Thx Ed, esp. for mentioning this...of course I got no notice from github on this, but ten from cri-o and openshift :confused:   Looking...

test_verbosefilter.py
```diff
@@ -0,0 +1,108 @@
...
+"""
+Unit tests for VerboseFilter, a logging helper in adept_openstack.py
+"""
+
+import logging
+import sys
+from mock import Mock
+from unittest2 import TestCase, main
+
+sys.modules['virtualenv'] = Mock()      # Needed for Ed's Fedora system
```
cevich: Yep, this is fine.  It's documented as a dependency, but  none of what you're testing here goes anywhere needing it, so ok for everyone, not just you :smile:

test_verbosefilter.py
```diff
@@ -0,0 +1,108 @@
...
+
+sys.path.insert(0, 'kommandir/bin')
```
cevich: also fine, it's never going to move.

test_verbosefilter.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class MyHandler(logging.Handler):
+    """
+    Output handler for logging messages. No output; just push onto list.
+    """
+    def __init__(self, message_list):
+        logging.Handler.__init__(self)
+        self.message_list = message_list
```
cevich: pylint may get mad (at some point) since there's no ``super`` call here.  The tests run under ubuntu, so god knows what the difference will be compared to Fedora.

test_verbosefilter.py
```diff
@@ -0,0 +1,108 @@
...
+    def setUp(self):
+        import adept_openstack  # pylint: disable=E0401
+        self.a_o = adept_openstack
+        self.logger = logging.getLogger()
+        self.message_list = []
+
+    def _test_basic(self, level, verbose, expect):
+        # Grrr. The Logger thing is a global, and there doesn't seem to be
+        # a way to get a new one each time. For addHandler() and addFilter()
+        # to work, we need to delve into internals.
```
cevich: Ya true...well, at least not without a whole lot of work.  Everything here is documented as python-2 (only) so this is fine since 2.7 is unlikely to ever change.

test_verbosefilter.py
```diff
@@ -0,0 +1,108 @@
...
+    (level, verbose, expect) = test_info
+
+    name = 'test_' + logging.getLevelName(level).lower()
+    if verbose:
+        name += '_verbose'
+
+    def _run_test(self):
+        self._test_basic(level, verbose, expect.strip()) # pylint: disable=W0212
+
+    return [name, _run_test]
```
cevich: Just for future reference, in ``unittest2`` and python3's ``unittest``, there is a ``subtest`` context manager that does this with less code.  It always runs all subtests, and shows the distinguishing features of any failures automatically.  Unf., the documentation is terrible.  It works like this:

```python
for bar in (thing, other_thing, third_thing):
    with unittest2.subtest(foo=bar):  # any kwargs here show up in failure messages, otherwise they're not used.
        result = some_method(foo)  # could just use bar directly if you wanted
        assertWhatever(result)
```

cevich: (what you have here is fine)

cevich:
FYI - it appears ``unit2`` isn't running this (for me or travis), investigating why...

test_verbosefilter.py
```diff
@@ -0,0 +1,108 @@
+# -*- python -*-
```
cevich: If I change the above line to ``#!/usr/bin/env python`` and make the file executable, it runs fine on it's own, but ``unit2`` doesn't seem to find (and run) it.  I'm willing to poke at it some more, or if it's okay with you, I'll just hard-code running this into ``.travis.yml``.

cevich: Oooohhh, I think I see the problem...rolling up a diff that should get this working with the ``unit2`` test-discovery...

edsantiago: I was not aware of the `unit2` command-line invocation. I've just been playing with it and the documentation is inaccurate and unhelpful. I also can't figure out why it's not picking up verbosefilter. I'll keep looking.

cevich: ...hmmm, nope.  How strange.  I'll have to sit and have a think-stink on the throne (or just hard-code it into ``.travis.yml``), since I promised mrunal work on the openshift installer :worried: 

Basically, running ``unit2`` should pick this suite up like it does all the others.  But I'm okay just hard-coding it too.

cevich:
> documentation is inaccurate and unhelpful

Wait, you found documentation? :smile: 

Don't waste time on it, really, the fact you wrote unit-tests is already above & beyond what I was expecting.  What I tried:

* Setting ``#!/usr/bin/env python``
* Making script executable
* Moving the generation-loop outside of the ``if __name__`` condition (so it runs on module import)
* Calling ``globals()`` and doing the setattr on the class from there.

I'm fine with hard-coding it, something like:

```diff
diff --git a/.travis.yml b/.travis.yml
index c0d12df..3335161 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ script:
     # Test building docs
     - cd docs && SPHINXOPTS="-W" make html
     - cd ..
+    - python test_verbosefilter.py
     - ./venv-cmd.sh unit2 -vfc
     - ./venv-cmd.sh ./test_exekutir_xn.sh
     # Check for some typos / git commit problems (except in this file)
```

(untested :smile:)

edsantiago:
Fixed and force-pushed. Reason: `unit2` was not seeing any `test_xxx` functions, because I was doing the test generation inside `if main`.

edsantiago:
Looks like it worked. Lines 869-876:
```
test_debug (test_verbosefilter.TestVerboseFilter) ... ok
[ and: test_debug_verbose, error, error_verbose, info, info_verbose, warning, warning_verbose ]
```

cevich:
> Fixed and force-pushed. Reason: unit2 was not seeing any test_xxx functions, because I was doing the test generation inside if main.

Hmmm, Interesting that was one of the things I tried.  Maybe I spelled something wrong :confused: Anyway, awesome, glad you got it working, hopefully it wasn't too much of a struggle.

cevich:
Thanks Ed!

PR #218, opened 2018-05-07
adept_openstack: refactor, allow expanded operations

edsantiago:
Remove much duplication and dangerous hardcoded knowledge
about the names and numbers of allowed operations. End
goal is to make built-in operations more modular, allowing
for simple expanding (or removing) of operations.

Also: clarified some confusing/misleading parts.

### DIFF
--- a/kommandir/bin/adept_openstack.py
+++ b/kommandir/bin/adept_openstack.py
@@ -22,6 +22,7 @@ import os
 import os.path
 import logging
 import random
+import re
 from importlib import import_module
 from imp import find_module
 import time
@@ -33,22 +34,22 @@ import shutil
 import virtualenv
 from flock import Flock
 
-# Operation is discovered by symlink name used to execute script
-ONLY_CREATE_NAME = 'openstack_exclusive_create.py'
-DISCOVER_CREATE_NAME = 'openstack_discover_create.py'
-DESTROY_NAME = 'openstack_destroy.py'
-REAP_NAME = 'openstack_reap.py'
-ALLOWED_NAMES = (DISCOVER_CREATE_NAME, DESTROY_NAME, ONLY_CREATE_NAME, REAP_NAME)
+# Operation is discovered by symlink name used to execute script,
+# e.g. 'openstack_exclusive_create'
+OPERATIONS = ('exclusive_create', 'discover_create', 'destroy', 'reap')
+
+# Stringified form: list of the valid names for invoking this script.
+ALLOWED_NAMES = ', '.join(['"openstack_{}"'.format(_) for _ in OPERATIONS])
 
 DESCRIPTION = ('Low dependency script called by ADEPT playbooks'
                ' to manage OpenStack VMs.')
 
 EPILOG = ('Required:  The WORKSPACE environment variable must exist and point to a'
           ' writable directory.  The script must be invoked by link (or symlink)'
-          ' named "%s", "%s", "%s", or "%s".'
-          ' If named "%s", VM creation will'
+          ' named %s'
+          ' If named "openstack_%s", VM creation will'
           ' fail if another with the same name already exist.'
-          % tuple(list(ALLOWED_NAMES) + [ONLY_CREATE_NAME]))
+          % (ALLOWED_NAMES, OPERATIONS[0]))
 
 # Needed for unittesting so argparse doesn't call system.exit()
 ENABLE_HELP = True
@@ -1066,7 +1067,7 @@ def parse_args(argv, operation='help'):
     Examine command line arguments, show usage info if inappropriate for operation
 
     :param argv: List of command-line arguments
-    :param operation: String of 'discover', 'create', 'destroy', 'reap', or 'help'
+    :param operation: one of the supported operations (see script), or "help"
     :returns: Dictionary of parsed command-line options
     """
     # Operate on a copy
@@ -1076,7 +1077,7 @@ def parse_args(argv, operation='help'):
                                      epilog=EPILOG,
                                      add_help=ENABLE_HELP)
 
-    if operation not in ('destroy', 'reap'):
+    if 'create' in operation:
         parser.add_argument('--image', '-i', default='CentOS-Cloud-7',
                             help=('If creating a VM, use this (existing) image instead'
                                   ' of "CentOS-Cloud-7" (Optional).'))
@@ -1096,10 +1097,10 @@ def parse_args(argv, operation='help'):
 
         parser.add_argument('--preserve', default=DEFAULT_PRESERVE, type=int,
                             help='The number of hours beyond one, before'
-                                 ' running "%s" would force deletion.'
-                                 ' Set to "0" for indefinite, otherwise defaults to'
+                                 ' running a reaper script would force deletion.'
+                                 ' Set to "-1" for indefinite, otherwise defaults to'
                                  ' %s (Optional).  Takes no action in --dry-run'
-                                 ' mode.' % (REAP_NAME, DEFAULT_PRESERVE))
+                                 ' mode.' % DEFAULT_PRESERVE)
 
         parser.add_argument('--size', '-s', default=None, type=int,
                             help=('If creating a VM, also create and attach'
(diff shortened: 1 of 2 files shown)


cevich:
Ohh, just found this.  Silly broken github notifications :confounded:

cevich:
Changes look fine on the surface to me, but travis is barking about some pylint stuff.  That may get fixed if you rebase and re-push.  Let me know if not, and/or you'd prefer not to to figure it out :smile: 

> Warning: completely untested

That's fine.  Adding integration tests back in is on a TODO list somewhere.  I can help test it manually (usually by running the 'basic' job).

edsantiago:
Rebased and force-pushed. I'm baffled by pylint failing because this module never passed pylint in the first place (as in, before I touched it).

cevich:
@edsantiago it's a "special" pylint built into the unittest, but yes, it barely passes.  I didn't investigate the issues it found here, perhaps a closer look would show that they can simply be ignored?

looking deeper...

cevich:
@edsantiago Oh, now I see it's passing pylint, but actual unittest is failing:

```

test_operations (test_openstack.TestParsedArgs)
Check expected minimum arguments parse in predictable way ... 
======================================================================
FAIL: test_operations (test_openstack.TestParsedArgs) (expected={'flavor': 'm1.medium', 'image': 'CentOS-Cloud-7', 'private': False, 'name': 'baz'}, operation='discover', parsed_args={'debug': False, 'operation': 'discover', 'verbose': False, 'timeout': 300, 'name': 'baz'})
Check expected minimum arguments parse in predictable way
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/cevich/ADEPT/test_openstack.py", line 203, in test_operations
    self.assertDictContainsSubset(op_expected[operation], parsed_args)
AssertionError: Missing: 'flavor','image','private'
----------------------------------------------------------------------
```

Should be easy to fix.

cevich:
BTW: U can reproduce exact same unittest run with:

``WORKSPACE=/tmp ./venv-cmd.sh unit2 -vfc``

edsantiago:
pylint and unit tests passing now - thanks.

cevich:
Yay!  Okay, I'll run the basic job against this...

cevich:
@edsantiago basic job appears to be working (so ``openstack_discover_create.py`` is fine), but I found a problem with the command-line parsing for destroy and reap:

```
./openstack_destroy.py --help
ERROR:root:Unknown operation 'destro'; script must be invoked with one of the following names: "openstack_exclusive_create", "openstack_discover_create", "openstack_destroy", "openstack_reap"
usage: ./openstack_destroy.py [-h] [--verbose] [--debug] [--timeout TIMEOUT]
                              name

Low dependency script called by ADEPT playbooks to manage OpenStack VMs.
...snip...
```

That error comes from around line 1368, and relies on ``parser.exit()`` in ``parse_args()`` with the special "help" operation.  The later part appears to be happening, but clearly somewhere the ``operation`` value is getting mangled :confused:   Any ideas?

kommandir/bin/adept_openstack.py
```diff
@@ -1357,18 +1356,17 @@ def activate_and_setup(namespace, venvdir, requirements, onlybin, nobin):
...
-        _dargs = parse_args(sys.argv, 'discover')
-    elif basename == ONLY_CREATE_NAME:
-        _dargs = parse_args(sys.argv, 'exclusive')
-    elif basename == DESTROY_NAME:
-        _dargs = parse_args(sys.argv, 'destroy')
-    elif basename == REAP_NAME:
-        _dargs = parse_args(sys.argv, 'reap')
-    else:
-        logging.error("Script was not called as %s, %s, or %s", *ALLOWED_NAMES)
+    basename = os.path.basename(sys.argv[0]).rstrip('.py')
```
cevich: somehow(?) this rstrip seems to be eating more than just the '.py' :confused:

cevich: my gosh that's odd...is ``rstrip()`` broken/bugged?

edsantiago: Yeah, python strip() doesn't seem to work the way I thought it did. Looking.

cevich: Oooohhhhh.....it's ``rstrip([CHARS])`` as in, characters to remove, not string.

cevich: This should work:

```python
    basename = os.path.basename(sys.argv[0]).rpartition('.py')[0]
    op_requested = basename.partition('openstack_')[-1]
```

cevich:
...basic job passed, but destruction both failed, and exited 0 :cry: That appears to be set here: https://github.com/cevich/ADEPT/pull/218/files#diff-d580606668133bebfaa51db413357274L1152

Changing that to ``parser.exit(1)`` results in non-zero exit, but the side-effect is then doing ``./openshift_whatever.py --help`` will also be non-zero exit (and maybe not a problem worth worrying about).

cevich:
Oh, I should clarify..."but destruction failed"...using the old code (with rstrip).

kommandir/bin/adept_openstack.py
```diff
@@ -1357,18 +1357,17 @@ def activate_and_setup(namespace, venvdir, requirements, onlybin, nobin):
...
-    elif basename == ONLY_CREATE_NAME:
-        _dargs = parse_args(sys.argv, 'exclusive')
-    elif basename == DESTROY_NAME:
-        _dargs = parse_args(sys.argv, 'destroy')
-    elif basename == REAP_NAME:
-        _dargs = parse_args(sys.argv, 'reap')
-    else:
-        logging.error("Script was not called as %s, %s, or %s", *ALLOWED_NAMES)
+    basename = re.sub(r'\.py$', '', os.path.basename(sys.argv[0]))
+    op_requested = re.sub('^openstack_', '', basename)
```
cevich: IDK in this instance, but for future ref: ``re`` is much more "expensive" than the simpler string manip. functions.

cevich:
...basic job running but my manual-testing all passes...

kommandir/bin/adept_openstack.py
edsantiago: New force push. Using `re`: `{,r}partition` seem close but could fail in (unlikely) corner cases. Might as well do it right. Tests pass.

kommandir/bin/adept_openstack.py
edsantiago: Python people are neurotic about re. Shrug. This is about maintainability.

cevich: Hmmm...not sure I agree...I think it's just me that's neurotic :smile:

cevich: gaah...github cross-comment-post.  What I mean is:  using 're' here is fine, we don't care about performance in this spot.   And you're right, it's more maintainable and less cryptic.

cevich:
...basic job setup passed, running DaT...

kommandir/bin/adept_openstack.py
edsantiago: Ha! No, I see it all the time when looking up python answers on StackOverflow. "re is inefficient", "re is unreadable", "re is evil". It's really weird, and counterproductive. I do agree that the Python re implementation is crap, as is pretty much all Python string handling, but the extent of the loathing for re is striking.

cevich:
...done, and all green :+1: 

Thanks @edsantiago, much cleaner code now :smile:

kommandir/bin/adept_openstack.py
cevich: > extent of the loathing for re is striking.

Interesting, I haven't encountered that.  Though I think you'll find such rantings about any language/system...people have opinions and love to spread them, especially when you disagree :smile:  But really the context definitely matters, for example right here, maintainability is _way_ more important than performance.  Best guess, it's got it's roots in 'C' (an original impetus and model for python). OTOH, try doing regex or printf stuff in the kernel, or for microcontroller and you'll be beaten to death with your own ripped-off limbs, then thrown into a volcano :laughing:

PR #220, opened 2018-05-23
Increase basic job peon preservation time

cevich:


cevich:
@edsantiago PTAL, and help testing always appreciated.

kommandir/inventory/group_vars/openstack/creation_destruction.yml
```diff
@@ -41,11 +41,12 @@ cloud_provisioning_command:
...
             {{ "--lockdir=" ~ hostvars.kommandir.global_lockdir | trim if hostvars.kommandir.global_lockdir|default() not in empty else ""}} \
-            --image {{ peon_image | trim }} \
-            --flavor {{ peon_flavor | trim }} \
+            --image={{ peon_image | trim }} \
+            --flavor={{ peon_flavor | trim }} \
             {{ "--private" if public_peons is undefined or public_peons != True else "" }} \
-            {{ "--size " ~ peon_size | trim if peon_size | default("",True) not in empty else "" }} \
-            {{ "--userdata " ~ peon_userdata | trim if peon_userdata is defined else "" }} \
+            {{ "--size=" ~ peon_size | int if peon_size | default("",True) not in empty else "" }} \
+            {{ "--userdata=" ~ peon_userdata | int if peon_userdata is defined else "" }} \
```
edsantiago: Why the change to `int`? In particular, I can't figure out what `peon_userdata` is supposed to be but as best I can tell it looks like a string? Unless I'm missing something major, command-line args are always strings, and it seems best not to try to second-guess what the type is supposed to be.

cevich: Nope, it was a typo, thx for pointing it out fixed.

cevich:
> And no, I can't think of any way to test it

Tested with ``bin/containertest.sh basic setup --headadept`` (as per docs I merged this week) then used openstack command to check the metadata.  Looks good.

cevich:
Having it set to 'int' in the template does extra validation, and trims it.  The entire command-line is being sent through a 'shell' task, so eventually it will all be string-converted anyway.  That's why the change.  It's never too late though,  if you think it should go back.  That's what version-control is for :smile:

PR #221, opened 2018-05-31
Increase default timeout for podman job

cevich:
I remember Ed mentioning this job takes 40-something minutes.  Fix the
default timeout to exceed that by a healthy margin.  Giving it lots
of breathing room is fine, since generally nobody is sitting around
waiting on an automated job.  I don't believe there are any shorter
timeouts in higher (execution) layers this will collide with.


cevich:
Manually tested.

PR #224, opened 2018-07-10
Fix TypeError from VerboseFilter

cevich:
Under some circumstances, modifying a logging message through a
logging filter causes false positives hits of the built-in
logging system's string-formating parameter checker.  Explored many
different workarounds and solutions for this, but none could pass the
verbose filter unittests.

Fix the problem by reducing the complexity of the overall script.
Replaced the filter with a dead-simple function. Then hand-separated
logging system messages from important operational details.  In
the later case, use the new function to direct them to the console.


edsantiago:
Do you have a reproducer for the TypeError? I don't understand what problem this is trying to solve.

cevich:
@edsantiago I don't have a ready-made one.  It's reproducing for me on F27 [after setting up the environment described in this document](https://url.corp.redhat.com/fde1a8e).

```
[cevich@cevich openstack]$ ./openstack_reap.py --dry-run --verbose
Traceback (most recent call last):
  File "/usr/lib64/python2.7/logging/__init__.py", line 868, in emit
    msg = self.format(record)
  File "/usr/lib64/python2.7/logging/__init__.py", line 741, in format
    return fmt.format(record)
  File "/usr/lib64/python2.7/logging/__init__.py", line 465, in format
    record.message = record.getMessage()
  File "/usr/lib64/python2.7/logging/__init__.py", line 329, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting
Logged from file openstack_reap.py, line 1294
Traceback (most recent call last):
...snip many more of the same...
```

I spent hours trying to fix/workaround using the filter and other code.  Anything in the filter code that alters the content/length of the message will trigger this problem.  That means any fix either adds lots more complexity, involves re-writing the unittests, or both.  Reverting 62e3c88b also fixes the problem, but breaks things in worse ways.

Not sure there are many good solutions here that doesn't involve hours and hours of more work :face_with_head_bandage: Just using a small function to print console messages is easy, doesn't need testing, and (should) work (still testing that).

cevich:
Closing in favor of https://github.com/cevich/ADEPT/pull/224

cevich:
errr...I mean #225

PR #225, opened 2018-07-10
VerboseFilter: handle format strings

edsantiago:
The last "fix" to VerboseFilter was broken: it would
replace a string like ("this is a %s", "test") with
("this is a test", "test") which would then yield:

    TypeError: not all arguments converted during string processing

Solution: don't replace the logging string. To handle
the previous issue (logger getting called with Exceptions),
in VerboseFilter, simply ignore non-string messages

### DIFF
--- a/kommandir/bin/adept_openstack.py
+++ b/kommandir/bin/adept_openstack.py
@@ -158,9 +158,9 @@ class VerboseFilter(logging.Filter):
         self.verbose = verbose
 
     def filter(self, record):
-        # record.msg may not be string; use getMessage to guarantee string form
-        msg = record.getMessage()
-        if msg.startswith('>'):
+        # record.msg may not be string, but '>' processing requires one
+        msg = record.msg
+        if isinstance(msg, str) and msg.startswith('>'):
             # Always remove leading '>'; this overrides the record string
             record.msg = msg.lstrip('>').lstrip()
             # When logging at INFO level (default), allow these only if verbose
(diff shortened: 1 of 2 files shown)


edsantiago:
@cevich in the interests of simpler diffs and of understanding the problem, how about this instead?

test_verbosefilter.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, message_list):
         self.message_list = message_list
 
     def emit(self, record):
-        self.message_list.append(record.msg)
+        self.message_list.append(record.getMessage())
```
edsantiago: This change is necessary to get `test_reformatting` to fail with the old code. It yields the same TypeError. With the new code, of course, it passes.

cevich:
> @cevich in the interests of simpler diffs and of understanding the problem, how about this instead?

Wonderful, and I'll happily accept it, except I tried something similar and it failed :confused: But lemmie take a closer look, maybe you did it differently/better...

kommandir/bin/adept_openstack.py
```diff
@@ -158,9 +158,9 @@ def __init__(self, logging_level, verbose):
         self.verbose = verbose
 
     def filter(self, record):
-        # record.msg may not be string; use getMessage to guarantee string form
-        msg = record.getMessage()
-        if msg.startswith('>'):
+        # record.msg may not be string, but '>' processing requires one
+        msg = record.msg
+        if isinstance(msg, str) and msg.startswith('>'):
```
cevich: Ya, this is essentially what I tried...but I won't rule out my having made some typo...

cevich:
...uggg, I really gotta get the ADEPT PR tester running again, testing these changes out by hand is getting old...

cevich:
...so far so good, reaper isn't dying...trying to run an ADEPT job using this HEAD...

